PR #1, opened 2023-09-21
nix flake

evanrichter:
Here's the start of a nix flake if you're interested in such a thing!

Currently it's useful for `nix develop` only, but I'm able to build and run kellnr on an M1 MacBook as well as an x86_64 Linux machine using this shell.

I'm experimenting with a new [flake builder library](https://github.com/accelbread/flakelight) and I think that's why `nix flake show` and `nix build` fail early. I expect `nix build` to not work because I haven't defined the `npm install` step at all, and I haven't tried any npm things in nix before so who knows how well that works.

Marked as a draft just to see if you're interested. Thanks for open sourcing your work!

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ resolver = "2"
 [workspace.package]
 authors = ["kellnr.io"]
 edition = "2021"
+name = "kellnr"
 version = "0.1.0"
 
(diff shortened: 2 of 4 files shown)


secana:
Hi @evanrichter, thank you for the draft.

Yes, I'm interested in a nix flake. Honestly I never used Nix, so my knowledge here is zero. Does it make sense to but the Nix flake in the kellnr repo itself, or is it more common to have a separate repo for that, e.g. like the Helm chart? I would like to have the most "common" approach.

PR #6, opened 2023-09-22
build: Fix invoking `npm` on Windows.

waywardmonkeys:


crates/kellnr/build.rs
```diff
@@ -5,6 +5,11 @@ static UI_DIR_SRC: &str = "../../ui/src";
 static UI_DIST_DIR: &str = "../../ui/dist";
 static STATIC_DIR: &str = "../../static";
 
+#[cfg(windows)]
+static NPM_CMD: &str = "npm.cmd";
```
secana, line 9: @waywardmonkeys why is that the case? When I run `npm` on Windows, it's just `npm` not `npm.cmd`.

secana (commented):
Waiting for an answer, why the change is needed

crates/kellnr/build.rs
waywardmonkeys: You probably run within a bash prompt and I was running it within a powershell prompt.

waywardmonkeys: Actually, if I use the git bash prompt within VS Code, I still get an error with the current code (not this PR):

```
$ cargo build
   Compiling kellnr v0.1.0 (C:\...\kellnr\crates\kellnr)
   Compiling registry v0.1.0 (C:\...\kellnr\crates\registry)
error: failed to run custom build command for `kellnr v0.1.0 (C:\...\kellnr\crates\kellnr)`

Caused by:
  process didn't exit successfully: `C:\...\kellnr\target\debug\build\kellnr-2580d729fbd88fee\build-script-build` (exit code: 101)
  --- stdout
  Build Kellnr - build.rs!
  cargo:rerun-if-changed=../../ui/src
  Installing node dependencies...

  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: NotFound, message: "program not found" }', crates\kellnr\build.rs:39:14
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
```

secana: I used powerhsell on Windows and npm was installed with winget. But it seems that 'npm' and 'npm.cmd' work, so it's better to use 'npm.cmd' if that works in anycase. Thx for the fix!

PR #50, opened 2023-10-02
Ported user routes to axum

ItsEthra:
I ported `/user` handles to axum without breaking frontend code. There is some room for improvements though, mainly
~~1.  Handles should return `Result<_, E>` where `E` is an `IntoResponse` type which logs important server errors, makes it easier to debug.~~
~~2. Didn't port any tests yet~~

So for now I'll leave this pr as draft

### DIFF
--- a/crates/web_ui/src/session.rs
+++ b/crates/web_ui/src/session.rs
@@ -91,12 +91,12 @@ impl MaybeUser {
 }
 
 #[axum::async_trait]
-impl axum::extract::FromRequestParts<appstate::ArcAppState> for MaybeUser {
+impl axum::extract::FromRequestParts<appstate::AppStateData> for MaybeUser {
     type Rejection = LoginError;
 
     async fn from_request_parts(
         parts: &mut Parts,
-        state: &appstate::ArcAppState,
+        state: &appstate::AppStateData,
     ) -> Result<Self, Self::Rejection> {
         let jar: CookieJar = parts.extract().await.unwrap();
         let session_cookie = jar.get(constants::COOKIE_SESSION_ID);
--- a/crates/web_ui/src/session.rs
+++ b/crates/web_ui/src/session.rs
@@ -3,7 +3,6 @@ use axum::http::request::Parts;
 use axum::RequestPartsExt;
 use axum_extra::extract::PrivateCookieJar;
 use db::DbProvider;
-use hyper::StatusCode;
 use rocket::http::{Cookie, Status};
 use rocket::outcome::Outcome::*;
 use rocket::request::{self, FromRequest, Request};
@@ -99,7 +98,7 @@ impl axum::extract::FromRequestParts<appstate::AppStateData> for MaybeUser {
                 Ok((name, true)) => Ok(Self::Admin(name)),
                 // not admin
                 Ok((name, false)) => Ok(Self::Normal(name)),
-                Err(_) => Err(RouteError::Status(StatusCode::UNAUTHORIZED)),
+                Err(_) => Err(RouteError::Status(axum::http::StatusCode::UNAUTHORIZED)),
             },
             None => Ok(Self::Guest),
         }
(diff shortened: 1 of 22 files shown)


secana:
Hi @ItsEthra thanks alot for the work. Escpecially seeing how the authentication is done in axum really helps.

I've created an issue to track the migration initiative: https://github.com/kellnr/kellnr/issues/51

crates/web_ui/src/session.rs
```diff
@@ -40,12 +43,73 @@ impl Name for AnyUser {
...
+        match self {
+            Self::Guest => None,
+            Self::Normal(name) | Self::Admin(name) => Some(name),
+        }
+    }
+
+    pub fn assert_atleast_normal(&self) -> Result<(), axum::http::StatusCode> {
+        matches!(self, MaybeUser::Normal(_) | MaybeUser::Admin(_))
+            .then_some(())
+            .ok_or(axum::http::StatusCode::FORBIDDEN)
```
jplatte: Just my 2c: I think this would be easier to read if it was a regular `match`.

crates/web_ui/src/session.rs
```diff
@@ -40,12 +43,73 @@ impl Name for AnyUser {
...
+
+#[axum::async_trait]
+impl axum::extract::FromRequestParts<appstate::ArcAppState> for MaybeUser {
+    type Rejection = LoginError;
+
+    async fn from_request_parts(
+        parts: &mut Parts,
+        state: &appstate::ArcAppState,
+    ) -> Result<Self, Self::Rejection> {
+        let jar = CookieJar::from_request_parts(parts, state).await.unwrap();
```
jplatte: With an extra `use axum::RequestPartsExt` this can be written as

```suggestion
        let jar = parts.extract::<CookieJar>().await.unwrap();
```

(I really don't like the `unwrap` there, but as `into_ok` is still unstable that's probably the best solution. I'll look into adding `extract_infallible` to `RequestPartsExt`)

jplatte: https://github.com/tokio-rs/axum/issues/2249

crates/web_ui/src/session.rs
ItsEthra: Fair enough, I'll change it

crates/web_ui/src/session.rs
ItsEthra: I'll use `RequestPartsExt` and leave unwrap then for now

ItsEthra:
Also I believe handles should use [PrivateCookieJar](https://docs.rs/axum-extra/latest/axum_extra/extract/cookie/struct.PrivateCookieJar.html), but it needs a key. afair rocket by itself manages cookie signing/encryption key, but axum requires key to be declared manually, so my question is how should user set it? Perhaps just an environment variable should be fine?

secana:
> Also I believe handles should use [PrivateCookieJar](https://docs.rs/axum-extra/latest/axum_extra/extract/cookie/struct.PrivateCookieJar.html), but it needs a key. afair rocket by itself manages cookie signing/encryption key, but axum requires key to be declared manually, so my question is how should user set it? Perhaps just an environment variable should be fine?
> 
> EDIT: I just realized there are settings, so maybe it's a better place for a key?

With rocket, a random key is generated on every startup for Kellnr. To keep it easy, we can take the same approach for axum. If Kellnr is restarted, the users need to login again, but that is perfectly fine as the session expires anyways.

There is the `common::util::generate_rand_string`  that can be used to get a random string.

ItsEthra:
Alright then, I'll implement it so the key is generated at startup and stored in the `AppState`.

crates/web_ui/src/user.rs
```diff
@@ -18,32 +19,51 @@ pub struct NewTokenResponse {
...
+            RouteError::DbError(err) => {
+                tracing::error!("Db: {err:?}");
+                StatusCode::INTERNAL_SERVER_ERROR.into_response()
+            }
+            RouteError::Status(status) => status.into_response(),
+            RouteError::InsufficientPrivileges => StatusCode::FORBIDDEN.into_response(),
+        }
+    }
+}
+
```
simonsan: I think something like this should go to a common `error.rs` for easier discovery?

ItsEthra: I'll move it to `error.rs` for now. It might be a good idea to have a common error type across crates but that would require a lot more refactoring.

ItsEthra:
@jplatte Do you have any pointers on how to do testing? I don't have experience with writing tests for web apps.

jplatte:
Does [this](https://github.com/tokio-rs/axum/blob/8854e660e9ab07404e5bb8e30b92311d3848de05/examples/testing/src/main.rs#L54C1-L54C1) help?

ItsEthra:
@secana I'm trying to port tests at the moment and I'm a bit confused about [this test](https://github.com/kellnr/kellnr/blob/main/crates/web_ui/src/session.rs#L320), does it ensure that admin cannot access pages meant for normal users? If so shouldn't admin be able access anything? The test expects not found code and It's just strange to me.

secana:
> @secana I'm trying to port tests at the moment and I'm a bit confused about [this test](https://github.com/kellnr/kellnr/blob/main/crates/web_ui/src/session.rs#L320), does it ensure that admin cannot access pages meant for normal users? If so shouldn't admin be able access anything? The test expects not found code and It's just strange to me.
> 
> EDIT: Similar question to the test `any_auth_user_but_no_cookie_in_store`, shouldn't anyone be allowed even if the cookie is missing?

There are three types of users in Kellnr.

1. The "normal" user. That's a user that is logged in but not an admin
2. The "admin" user. Logged in user with more rights
3. The "any user" that is a "normal" or "admin" user but still logged in.

There are pages that are only accessible to either one of them. For example the normal user sees a different settings page than the admin user. Some pages are the same for normal and admin users. That's where the "any user" is used.

The `normal_auth_user_is_admin` checks is, if a route which is for a normal user is requested by an admin user, a 404 is returned. Concrete example: The admin user should not get the normal user settings.

The `any_auth_user_but_no_cookie_in_store` tests what a logged out user should see. If any user tries to authenticate with a session token, but the token is not known by Kellnr (because it expired or the db was reset), "unauthorized" has to be returned, such that the user can login again.

I hope that helps.

ItsEthra:
I see, thanks, that helps.

ItsEthra:
Alright, I believe I ported most tests and all routes under `/user` path. I'm still not very happy about the way I set cookies in tests but I haven't found a better way to do so.

secana:
I'm on a business trip for the next few days. I'll review and merge when I'm back. So far: thank you for the incredible work.

Issue #93, opened 2023-10-29
Authentication on pull must require auth in UI


secana:
Currently, if the "authentication on pull" feature is enabled, any user (logged in or not) can see the crates, docs and so on in the UI. That renders the authentication feature rather useless, as the code of the crates can be seen in the docs anyway.

Implement: If the "authentication on pull" is enabled, require a login to the UI and do not show any data to anonymous users.

secana:
Next to-do: Check in the UI if auth if required and if so, show only the login page, instead of an empty list of crates.

PR #105, opened 2023-11-03
Add a middleware for Authentication of docs route.

sinyo-matu:
Hi @secana!

This is in relation to issue #93.

## What this PR does
- Adds a middleware `auth_when_required` to `web_ui::session` and sets the `session` module to public.
- Utilizes the middleware `auth_when_required` to protect the `docs/` route that serves documentations.

### What `auth_when_required` does
It checks if a user is logged in when `settings.auth_required` is set to true.
- If the user is logged in, it allows the request to pass through.
- Otherwise, a 401 status code is returned.

With these changes, only logged-in users can access the documentation when `settings.auth_required` is set to `true`.

### DIFF
--- a/crates/kellnr/src/main.rs
+++ b/crates/kellnr/src/main.rs
@@ -1,5 +1,6 @@
 use appstate::AppStateData;
 use axum::{
+    middleware,
     routing::{delete, get, get_service, post, put},
     Router,
 };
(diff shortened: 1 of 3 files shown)


secana:
Thanks a lot! I'll use the middleware approach in other parts of Kellnr, too. Much better than adding a not used parameter to each function, just to get authentication enforced.

sinyo-matu:
> Thanks a lot! I'll use the middleware approach in other parts of Kellnr, too. Much better than adding a not used parameter to each function, just to get authentication enforced.

@secana Thank you for merging this!
I'm totally agree with you.
And it also reduce unnecessary dependency which is really satisfying.

PR #140, opened 2023-11-28
Migrated to axum 0.7

ItsEthra:
We just updated from rocket to axum and now we need to update to axum 0.7 🥲 oh well. ~~Right now mostly blocked by reqwest~~

### DIFF
--- a/crates/auth/src/auth_req_token.rs
+++ b/crates/auth/src/auth_req_token.rs
@@ -204,10 +204,6 @@ mod auth_middleware_tests {
     use std::sync::Arc;
     use tower::ServiceExt;
 
-    async fn handler_return_200() -> StatusCode {
-        StatusCode::OK
-    }
-
     async fn app_required_auth(db: Arc<dyn DbProvider>) -> Router {
         let settings = Settings::default();
         let state = AppStateData {
@@ -237,7 +233,7 @@ mod auth_middleware_tests {
             ..appstate::test_state().await
         };
         Router::new()
-            .route("/guarded", get(handler_return_200))
+            .route("/guarded", get(StatusCode::OK))
             .route_layer(from_fn_with_state(state.clone(), cargo_auth_when_required))
             .with_state(state)
     }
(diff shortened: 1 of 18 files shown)


ItsEthra:
Workflow must be fixed to not run on draft prs I believe

secana:
Wow, you are fast :)

The pipeline runs only the tests. Build/Deploy is skipped and only run if a "Github Release" is triggered. Thanks again, for putting in the work!

ItsEthra:
Alright thanks, sorry for confusion I don't have much experience with github actions.

crates/auth/src/auth_req_token.rs
```diff
@@ -220,10 +221,11 @@ mod auth_middleware_tests {
             }),
             ..appstate::test_state().await
         };
+
         Router::new()
-            .route("/guarded", get(handler_return_200))
+            .route("/guarded", get(StatusCode::OK))
```
jplatte, line 222: Nice to see the handler impl for `T: IntoResponse + Clone` get some use! Any reason you didn't replace the last remaining use of `handler_return_200` as well?

crates/web_ui/src/user.rs
```diff
@@ -175,8 +178,8 @@ pub async fn logout(
         None => return Ok(jar), // Already logged out as no cookie can be found
     };
 
-    jar = jar.remove(Cookie::named(COOKIE_SESSION_ID));
-    jar = jar.remove(Cookie::build(COOKIE_SESSION_USER, "").path("/").finish());
+    jar = jar.remove(Cookie::build(COOKIE_SESSION_ID));
```
jplatte:
```suggestion
    jar = jar.remove(COOKIE_SESSION_ID);
```

This should be the same, right?

crates/auth/src/auth_req_token.rs
ItsEthra: I was just trying to fix the errors, I guess it's a good idea to replace other uses of it, will do.

PR #185, opened 2024-01-30
fix: max crate size limit

### LINKED ISSUE #184: Crates upload body limit
crates upload body limit is not set in axum, the setting `crates/max_crate_size` in `default.toml` is not working

dupeiran001:
fix #184 by add upload body limit to axum


secana:
Thank you for the fix. It will be in the next release in a short time.

PR #220, opened 2024-03-08
Minor improvements

neon-JS:
improvement: Send 401 instead of 500 on invalid login credentials

----

fix: Make page scrollable on small screens

----


neon-JS:
Those two things caught my eye when setting up kellnr. I'm new to this project, so I don't really know how you handle contributions. If there's a contribution guide, feel free to drop a link :)

Issue #254, opened 2024-03-26
Doc build in kellnr docker container fails with: Invalid cross-device link


mbodmer:
What could be the cause for failing doc builds inside the container?

`Invalid cross-device link (os error 18)`, while trying to rename 
from `/root/.rustup/toolchains/...` to `/root/.rustup/tmp/...` 
does not make sense, because this is the same filesystem as I understand it.

Any ideas?

Thanks for kellnr, by the way. It's great!

```sh
Running `rustc --crate-name regex --edition=2021 
/root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-1.10.4/src/lib.rs 
--error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat 
--crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no 
--cfg 'feature="std"' --cfg 'feature="unicode-perl"' 
-C metadata=acec26d0b77f6756 -C extra-filename=-acec26d0b77f6756 
--out-dir /opt/kdata/doc_queue/cRvqR6YjeT/mylib-sys-2.4.0/target/debug/deps 
-L dependency=/opt/kdata/doc_queue/cRvqR6YjeT/mylib-sys-2.4.0/target/debug/deps 
--extern regex_automata=/opt/kdata/doc_queue/cRvqR6YjeT/mylib-sys-2.4.0/target/debug/deps/libregex_automata-f6805926912f666c.rmeta 
--extern regex_syntax=/opt/kdata/doc_queue/cRvqR6YjeT/mylib-sys-2.4.0/target/debug/deps/libregex_syntax-5957eeb6d1504fb2.rmeta --cap-lints allow`
info: removing previous version of component 'cargo'
info: rolling back changes
error: could not rename component file from 
'/root/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/cargo' to
'/root/.rustup/tmp/xrl8fy9awgv8lkcy_dir/bk': Invalid cross-device link (os error 18)
error: could not compile `proc-macro2` (build script)
```

secana:
Hi. Thanks for reporting. 
Does this only happen with this specific crate? Is there anything special (other toolchain, edition)? 
I've never seen this error before, I need more information to replicate it.

mbodmer:
Hi, thanks for the fast reply.

While researching the error I found this issue: https://github.com/rust-lang/rustup/issues/1239

[...]

PR #255, opened 2024-03-28
dockerfile: add additional dependencies for sys-crate doc builds

mbodmer:
- libclang1: bindgen needs libclang,
  see: https://github.com/rust-lang/rust-bindgen
- cmake: is needed by many native libs

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,7 @@ ARG AUTHTOKEN=Zy9HhJ02RJmg0GCrgLfaCVfU6IwDfhXD
 ARG VERSION
 
 RUN apt update
-RUN apt install -y curl sed git zip apt-utils build-essential libclang1 cmake
+RUN apt install -y curl sed git zip apt-utils build-essential llvm-dev libclang-dev clang cmake
 
 # Install rust for the automatic rustdoc generation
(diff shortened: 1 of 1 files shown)


syphar:
I understand that the doc build is executed inside the kellnr docker container, so these dependencies are only needed to build docs hosted on kellnr? 

This will only become more and more. 

docs.rs and crater are using [our own `crates-build-env` container](https://github.com/rust-lang/crates-build-env), see also https://docs.rs/about/builds, and the [list of installed packages is long](https://github.com/rust-lang/crates-build-env/blob/master/linux/packages.txt).

mbodmer:
Thank you for the pointer, I did not know this. I think it would make sense to use this container as the base. The only thing is it would probably trigger this error again with Ubuntu 22.04: https://github.com/kellnr/kellnr/issues/254

mbodmer:
Thinking a 2nd time about it, I am not 100% sure it makes sense to carry that much dependencies for everybody. I see 3 groups of sys dependencies:

  1. basics which enable many sys builds: build-essential, cmake, meson, clang, ...
  2. often used but debatable: protoc, flatc, smithy, ...
  3. out of scope: others

One can easily extend the docker container with dependencies by inheriting from the kellnr container. This is much less convenient, however.

PR #257, opened 2024-03-28
build docs with all features enabled

mbodmer:
My docs won't have much content when not built with --all-features.
I think this would be a sensible addition for everyone.


secana:
Hi @mbodmer! Thank you for your PRs. Unfortunately this one fails a test. Can you have a look?

mbodmer:
Hi, i hope it's better now. Unfortunately building is hard on my system (nodejs...).

PR #295, opened 2024-04-16
Share cache between prefetch tasks

Christiaan676:
Hi, noticed that the cache is not shared between the crates_io prefetch tasks.  And made a draft to changes this, but as it requires moving some settings up to main, I wonder if it would not be better to spawn the tasks in `cratesio_prefetch_api.rs`. Let me know what you think is best and I'll update that.

Als removed the Arc around the flume channels. They already contain an Arc internal, so wrapping them in an Arc is a bit double.

### DIFF
--- a/crates/kellnr/src/main.rs
+++ b/crates/kellnr/src/main.rs
@@ -9,7 +9,7 @@ use axum_extra::extract::cookie::Key;
 use common::cratesio_prefetch_msg::CratesioPrefetchMsg;
 use db::{ConString, Database, DbProvider, PgConString, SqliteConString};
 use index::{
-    cratesio_prefetch_api::{self, background_update_thread, cratesio_prefetch_thread},
+    cratesio_prefetch_api::{self, init_cratesio_prefetch_thread},
     kellnr_prefetch_api,
 };
 use registry::{cratesio_api, kellnr_api};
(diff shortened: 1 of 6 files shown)


crates/kellnr/src/main.rs
```diff
@@ -207,8 +207,14 @@ async fn init_cratesio_prefetch_thread(
...
+
+        static UPDATE_CACHE_TIMEOUT_SECS: u64 = 60 * 30; // 30 min cache timeout
+
+        let cache = Cache::builder()
+            .time_to_live(Duration::from_secs(UPDATE_CACHE_TIMEOUT_SECS))
+            .build();
         tokio::spawn(async move {
-            cratesio_prefetch_thread(db2, recv2).await;
+            cratesio_prefetch_thread(db2, cache.clone(), recv2).await;
         });
```
Christiaan676: Don't really like this, and like I mentioned can move the init code to the cratesio_prefetch module.

secana:
Hi @Christiaan676. First: Thanks for contributing and improving Kellnr. I think you are right and a cleaner solution is to move the code to the cratesio_prefetch module. It would be great if you could refactor the code.

Christiaan676:
@secana Made the changes.

secana (approved):

PR #366, opened 2024-06-20
Fix zip version and rework docker image build

MarcAntoine-Arnaud:


secana:
Thanks for the PR, but I have multiple issues with it:

- It switches from Ubuntu to Debian, which I would like to prevent, as all tests are run for Ubuntu and I don't want to provide support for two distributions
- It rebuilds Kellnr, which is already build in the Github pipeline. By using the prebuild binaries, it's guaranteed that Kellnr in the container image and Kellnr for bare-metal are the same. 
- The `start.sh` is missing. It's crucial, as it imports the certificates needed to correctly setup TLS for Kellnr

Regarding https://github.com/rust-lang/rustup/issues/2949, I would prefer to check, how the official Rust image prevents the issue and port that to Kellnr.

What are your thoughts?

MarcAntoine-Arnaud:
Hi @secana,

Here my replies:
- about Ubuntu/Debian, I think Debian is more appreciated in term of security, and Kellnr is quite important in term of security. And Rust handle multi-os very well, and Debian and Ubuntu are so closed. I'm not sure you have to take care about it. Maybe we can add Debian in the CI if you prefers, but for me it's quite useless.
- I'm not sure to ensure the second point. For me I found to rebuild into the docker (using build image, than copy executable into runtime image) is the best in term of compatibility. For me distribute for bare metal and docker is 2 different use cases. And as new developper, I found that way less understandable (with the VERSION parameter, it means you can build a docker image with any version of Kellnr, so in term of versioning it's not safe for me)
- about certificates, we can document to mount a volume to the right place with certificates. So I didn't see the reason to keep the start script.

Feel free to re-comment my remarks, and if it's simplest we may discuss by visio this week.

secana:
Hi,

I'm sold on the switch to Debian. But I think, I didn't understand the issue with the version. To release a new version of kellnr, a tagged git release e.g. `git tag -a vX.X.X` triggers the workflow that builds kellnr and releases the binaries to github release. These releases are than taken to build a docker image with the same version. To be more precise with the tags `vX.X.X`, `vX.X` and `vX`. 
What would you like to change about that?

Q: How do you mount certificates at runtime into the container? I don't like that a `start.sh` script is needed just to import the certificates, but I could not find another way to achieve a reliably import of certificates.

MarcAntoine-Arnaud:
Hi,

Cool for Debian :D 
For version, it's about Cargo.toml (https://github.com/kellnr/kellnr/blob/main/crates/kellnr/Cargo.toml#L5) it's in `0.1.0` and the tag in 5.0.0. using tool like `cargo release` will update you all `Cargo.toml`, create a conventionnal commit, and add a tag (we use that for as internal process).

With docker, you can mount anywhere a volume (https://docs.docker.com/storage/bind-mounts/). So anyone can mount into `/usr/local/share/ca-certificates/` any volume with it's data. By adding it in the documentation, no script is required. And I think it's relevant (for security and operations reasons) to not build image with included certificates.

secana:
Mounting a certificate is documented here: https://kellnr.io/documentation#rustdoc but it only works because the `start.sh` runs `update-ca-certificates` before kellnr is started in the container.
I tried a simple example with mounting a certificate into an empty container and a `curl` to a webpage that requires the cert and the request failed, as the cert was not imported with `update-ca-certificates`. 

Did I miss something and it's possible to import a certificate at runtime without running `update-ca-certificates`? Getting rid of the `start.sh` would be great, but it seems to me that the minimum required is a `update-ca-certificates && kellnr`.

MarcAntoine-Arnaud:
I see, for you in that case you have to redefine the command (CMD).
2 options are available:
1. update the Dockerfile and do: `CMD update-ca-certificates && kellnr`
2. let the user to do it, `docker run [OPTIONS] IMAGE [COMMAND] [ARG...]`, means we can run: `docker run -v ./certs:/usr/local/share/ca-certificates/ kellnr "update-ca-certificates && kellnr"`

Option one is quite the most elegant for end-users, the second is more adapted to specific use cases (kellnr may have other custom configuration like that and may provide a long start-time).

secana:
I'll close this PR, as I need to rework the whole CI pipeline to switch to Debian without introducing a breaking change in binary compatibility with ubuntu.

Tests on the devel branch have shown that building in docker takes 4.5h on Github instead of the 15 minutes, if kellnr is build outside of a container. Even if the build split to multiple runners with a few Docker tricks (https://docs.docker.com/build/ci/github-actions/multi-platform/), it would take ~1.5h to build.

The long build times are due to how `docker buildx` for multi-architecture images works. As the CI runner is x86_64, the ARMv7 and AARCH64 are emulated with qemu, which takes forever. On the other hand cross-compiling for a different architecture likes the current approach does it, has no performance drawbacks.

The most promising way seems to be to build a static binary with musl that can be run on any linux and copy that to the Debian image like it's done with the Ubuntu image currently.
This will not break binary compatibility with Ubuntu and the build times are acceptable. One more advantage is that Kellnr can run on any distribution, not Ubuntu exclusively.

secana:
Kellnr 3.2.4 is now based on the official rust image. I'll close this PR.

secana:
Kellnr 3.2.4 is released and based on the official rust image. Thanks for the push in that direction!

Issue #209, opened 2024-02-22
Permission to push and pull


vikulikov:
It would be great to be able to set permissions to push and pull for every user for every crate or for group of users and group of crates

secana:
Hi @vikulikov,
for push that is already possible. Each crate as one or more owners, which are allowed to push new versions. Pull is allowed for every (authenticated) user. That replicates how crates.io works.

What use-case do you have, where pull needs to be per user?

vikulikov:
We have a lot of internal utility crates with different permission levels for every user. So there are crates that are public for every user in the company and there are that are restricted to some groups. And it would be convinient to manage these permissions and all internal crates through one registry instance

secana:
Hi! I released a new version of kellnr 5.3.0-rc1. If you have time, it would be great if you could check if everything works as expected. If so, I can release the official 5.3.0 version.

PR #377, opened 2024-07-05
Feature #209: initial implementation of download authorization

jauhien:
This is an implementation of crate download authorization for issue #209.

In our use-case we need to restrict download of crates on per user basis (we intend to use kellnr in our private crate registry in one of the projects).

Regarding the API design I tried to follow the HTTP semantics, so PUT/DELETE operate on a given user.

Also we would need more changes regarding user management and web-hooks that I would be happy to discuss them with you.


secana:
Hi @jauhien, 
sry for the late response. I had some busy weeks. First: Thanks for putting in that much work!

What do we need to discuss to get the feature working an merged?

jauhien:
Hi @secana , the feature concerned in this MR as implemented can be merged and is working.

We would need other changes, but for now we are focused on other work, so I do not expect that we will be able to discuss them before September.

Regarding this MR I would be grateful if you could do code-review, so I can make changes based on your feed-back (if any) and the final version can be merged by you.

secana:
A question for clarification: 
Is the introduction of a `crate_user` necessary? Wouldn't it be much easier to use the existing concept of a an `owner` and enhance it, such that it has a flag for "read" or "read/write" access? It seems to me that this would do the same with half the code.
Or do I miss something and the introduction of a `crate_user` relation has some more benefits?

jauhien:
Sorry for a long pause: I was on vacations and then had other topics to manage.

Your comment sounds reasonable to me, so I will redo those changes and update PR.

jauhien:
Dear @secana , I have thought a bit more and there is one thing that probably makes `crate_user` necessary: we do not want to allow crate owner with write access to be able to change crate users list as in our use case crates uploading is separated from granting access to crates.

In current implementation crate owners can change list of owners (which seems to be reasonable as there can be a necessity to pass upload authorization to other user or to revoke it).

So I can see two options: keep separated crate user entity or restrict crate owners changes to admins only.

Which option seems better to you?

vikulikov:
@secana Hello! Is there something that is blocking this PR from merge?

secana:
Hi, I've reviewed most of it and think it's ready to be merged. As it is a big new feature, would you be so kind and add documentation about it for https://kellnr.io? The correspondig page for docs is here: https://github.com/kellnr/website/blob/main/src/views/DocumentationV5View.vue

jauhien:
@secana , thank you very much!

Surely, I can add the docs. I will have a look this week.

jauhien:
@secana , sorry for a longer pause, I have created a pull request on the website repo.

Please, do not hesitate to comment if you need more documentation.

From the side of user this feature is seen as an additional settings tab (available to administrator) and as a mechanism that rejects cargo pulls for unauthorized users, so documentation is not long.

PR #535, opened 2024-12-12
Make version, as well as config and static directories configurable

gtsiam:
I'm trying to package kellnr [for the AUR](https://aur.archlinux.org/packages/kellnr), but the current model of dumping everything in the working directory will not work for me. I need the option to compile kellnr to use sensible paths (`/etc/kellnr` and `/usr/share/kellnr`) for its accompanying files. [^1]

This pull requests adds the `KELLNR_CONFIG_DIR` and `KELLNR_STATIC_DIR` compile-time environment variables. If these are not set, behaviour is unchanged [^2]

EDIT: Also UI version was apparently introduced by a sed on CI. Let's not do that.

<details>

<summary>Sidenote</summary

Personal opinion: Merging with rebase exists, use it. Merge commits everywhere create spaghetti commit history. Don't get me wrong, merge commits have their place (large merges / complicated merge conflicts). But they shouldn't be used by default for small-scale projects.

Feel free to ignore this part.

</details>

[^1]: Yes symlinks might work, but I didn't even try. This makes a lot more sense imho.

[^2]: Well, technically main now uses `setting::get_settings()` which by default looks up some directories for the config dir. So that changed, but I don't see it being a problem.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,9 +80,6 @@ jobs:
           echo "KELLNR_VERSION=${GITHUB_REF#refs/*/v}" >> $GITHUB_ENV
           echo $KELLNR_VERSION
 
-      - name: Replace Version in web_ui/ui
-        run: sed -i 's/0.0.0-debug/'"$KELLNR_VERSION"'/' crates/web_ui/src/ui.rs
-
       - name: Build Release {{ matrix.target }}
         run: just target=${{ matrix.target }} ci-release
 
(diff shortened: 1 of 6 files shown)


gtsiam:
I also started randomly getting test failures. Turns out, migrations cannot run without a user-writable data dir (this is a problem for tests, since you'd assume a data dir doesn't really exist for testing).

I just fixed the relevant migration for now.

gtsiam:
Nope, it's good. Go for it!

PR #576, opened 2025-01-17
Add endpoint to get all versions of a crate

CarlSchererKenbun:
Hey, I added an endpoint, `/api/v1/crates/:crate_name/crate_versions`, that returns all versions of `crate_name`.

My reason for wanting this is that in the CI I want to check if the version of my crate already exists in the registry and only publish if it hasn't already been released.

`cargo search` or `cargo info` only provide information about the latest version, which is sufficient most of the time, but when working on a fix for an older version, getting the latest version is not enough.

Feel free to change this however you want or just tell me what needs to be changed or added.

### DIFF
--- a/crates/db/src/database.rs
+++ b/crates/db/src/database.rs
@@ -949,6 +949,18 @@ impl DbProvider for Database {
             .collect())
     }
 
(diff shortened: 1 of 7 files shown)


secana (approved):

secana:
@CarlSchererKenbun thanks for the PR! The new API will be in the next release coming soon.

Issue #173, opened 2024-01-16
Support S3 as file storage


Flowneee:
Hi. I wanted to ask about S3 support - is it planned? I think it could be very useful in environments where it is available. For example in my current environment I have no access to any kind of persistent storage unfortunately, so deploy in k8s is out of question. For example https://github.com/mcorbin/meuse supports it (we currently use it, but it is lack certain features like docs generation and user management in UI, also written in Clojure so hacking it is much harder).

If you ok with it, I will try to implement it and send a PR.

secana:
Hi,

yes it's planned. When I built kellnr, I abstracted the storage as far as possible, such that a switch of the storage provider is possible. I used the same to support multiple database backends. I don't know when I have time to tackle the S3 backend, but I'll keep this issue as a reminder.

MarcAntoine-Arnaud:
It may relevant to use OpenDAL, so it will abstract various storages, inc. S3 and local and Google Storage etc.

secana:
@MarcAntoine-Arnaud Thanks for the idea. I never heard of OpenDAL before, but it looks very promising.

MarcAntoine-Arnaud:
Yes we use it for an internal project and it's cool. 
We may help to integrate it.

Omega359:
Another solution could be to use [object_store](https://docs.rs/object_store/latest/object_store/).

kushudai:
I took a first hacky stab at this in https://github.com/kellnr/kellnr/pull/600
Feedback is appreciated. It's certainly _not_ ready for a review with the intention of merging that code.

secana:
Hi @kushudai! Thank you for the PR. While it's quite hacky it show that a S3 integration does not seem to be too much work. Are you interested in improving the PR futher?

kushudai:
> Hi [@kushudai](https://github.com/kushudai)! Thank you for the PR. While it's quite hacky it show that a S3 integration does not seem to be too much work. Are you interested in improving the PR futher?

Absolutely, I would appreciate some advice on how we want to deal with the Drop trait requirement on the S3File type :)

secana:
I'll have a look as soon as I've time. I never used S3 from Rust, so that's a new one for me.

secana:
In progress in PR #623

PR #600, opened 2025-02-04
Hacky S3 support

kushudai:
Just trying to figure out what it would take to support S3 as a backend instead of using local disks.

Just a first stab at addressing https://github.com/kellnr/kellnr/issues/173


crates/storage/src/cached_crate_storage.rs
```diff
@@ -110,19 +110,19 @@ impl CachedCrateStorage {
...
+        }
+    }
+
+    pub fn cache_has_path(&self, file_path: &PathBuf) -> bool {
+        if let Some(cache) = &self.cache {
+            cache.contains_key(file_path)
+        } else {
+            false
+        }
+    }
```
kushudai, line 125: Apologies, I ran cargo fmt

crates/storage/src/storage_provider.rs
```diff
@@ -2,26 +2,30 @@ use axum::async_trait;
...
-use crate::storage_error::StorageError;
-
 #[async_trait]
 pub trait StorageProvider: Send + Sync {
-    async fn open_file(&self, file_path: &Path) -> Result<File, StorageError>;
-    async fn open_or_create_file(&self, file_path: &Path) -> Result<File, StorageError>;
-    async fn read_file(&self, file: &mut File) -> Result<String, StorageError>;
-    async fn read(&self, path: &Path) -> Result<Vec<u8>, StorageError>;
+    type Err;
+    type StoredFile;
```
JanBerktold, line 8: I don't think this will work w/o any traits -- how will consuming code know how to modify the file?

crates/storage/src/s3_storage.rs
```diff
@@ -0,0 +1,146 @@
...
+
+pub struct S3File {
+    pub path: PathBuf,
+    pub content: Bytes,
+    s3: Arc<S3Storage>,
+}
+
+/// This is quite horrific but the StorageProvider trait forces this on us as it gets a file 
+/// back and can do whatever it wants with said file without explictly committing changes.
+impl Drop for S3File {
```
JanBerktold, line 31: Can we introduce a trait for StoredFile with an explicit drop?

crates/storage/src/storage.rs
```diff
@@ -23,6 +23,8 @@ impl Default for Storage {
 
 #[async_trait]
 impl StorageProvider for Storage {
```
JanBerktold, line 25: Can we add config to actually use the new S3 provider & wire it up? I think that will reveal some more problems.

secana:
There is a more advanded implementation PR here: #607
I'll close this one in favour of #607.

PR #607, opened 2025-02-16
Feature/s3 support

homycdev:
in this PR: 
- removed raw tokio file interraction. 
- introduced Storage type that interacts with AmazonS3 and LocalFS 
- added integration tests for s3(minio) storage, fixed existing tests that work with local file system
- refactored cratesio_api.rs


secana:
Hi. Thank you for putting in the effort. I really appreciate that!

I know that the following request is some work, but I would prefer it, if instead of addin the S3 storage with an "if" to the existing storage classes, it gets abstracted into a trait. E.g. some `AbstractStorage` trait that is either a filesytem or a S3. 

IMHO this is a cleaner design as it keeps the code for a filesystem backend and s3 backen separate, which is easier to read and test. Furthermore it allows easy addition of new backends in the future, as just the trait has to be implemented. 

Could have a look into that idea?

homycdev:
sure, no problem

homycdev:
do you see your StorageProvider trait as a final design, or is there room to discuss it?

secana:
The `StorageProvider` is a left over and already removed in the `devel` branch of kellnr. So I guess now is the time for a fresh start with a storage trait that abstracts the behavior. As the `devel` branch is ahead of the `main` branch with a few fixes, it may be a good idea to base this PR against the `devel` branch.

homycdev:
closed, due to opened PR to dev branch https://github.com/kellnr/kellnr/pull/623

PR #625, opened 2025-02-27
Add read-only users

### LINKED ISSUE #601: Feature request: read-only user
We have a handful of internal crates that are currently being managed as Git dependencies, which is a bit unwieldy and requires having SSH keys in CI runs to download dependencies. Upgrading is also a pain since it requires looking through the Git log for each dependency and choosing a compatible commit (some with breaking changes).

So we've decided to set up a Kellnr registry.

The ideal workflow in my head is to check-in a registry definition with an auth token to each project. That way, if someone has access to the repository, they can pull it and its dependencies and build without any extra fuss. Ideally, I'd like this token to only be able to _download_ crates, not publish them.

I've created a `pull-only` user to do this, and made sure they weren't a user of any crate, but I don't see any way to actually restrict the ability to publish new crates to the registry. If the auth token were to leak somehow, an attacker _could_ pull all our source code, which wouldn't be great but it likely means that they had access to it anyway; but they could also publish junk crates to the registry and fill up the `/opt/kdata` volume, or use it to deliver some sort of attack payload and chain exploits.

I'd additionally like to be able to restrict this user to only downloading certain crates, that way we could create pull tokens for potential partners without giving them access to everything in the registry. But I can't do that right now without also making the user an owner of the crate, and thus giving it the ability to publish new versions.

5Dev24:
Resolves #601 

* Updated existing tests to have read-only set to false, better choices should be considered
* Added `is_read_only` cached field to `Token`
* Created migration to add read-only column and updated entities, default is false for existing rows
* Added `is_read_only` column to User model
* Added `DBProvider` method to change a user's read-only state
* Added `/read_only/:name` endpoint for updating a user's read-only state
* Block read-only users from removing owners, removing crate user's, adding owners, adding crate user's, publishing, yanking, and unyanking
* Fixed issue with username's containing path characters in web ui `remote-routes.ts`
* Display button to switch a user's read-only status in the admin UserMngr panel
* Added `is_read_only` field when creating a new user in admin UserMngr panel

Sorry for another large PR so soon.

Please check that my migrations work for existing instances of Kellnr for both SQLite and PostgreSQL. It worked locally for migrating both backends, but getting another developer to verify would be appreciated.

Please check that the web UI works well and looks good. I'm fine with changing the text and/or style of the User Management button I added. I tried to integrate the same feel as the other buttons there.


crates/db/migration/README.md
```diff
@@ -2,7 +2,7 @@
 
 - Generate a new migration file
     ```sh
-    cargo run -- migrate generate MIGRATION_NAME
+    cargo run -- generate MIGRATION_NAME
```
secana, line 5: Did that change with a newer seaorm version?

secana (approved):

secana:
Thanks again for such a great PR! I've reviewd it and it looks good. I'll merge it into devel and give the migrations a spin.

crates/db/migration/README.md
5Dev24: Yeah, the command has changed. Got it from the docs [here](https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli/). The README on the cli's crate hasn't been updated in over 3 years

PR #671, opened 2025-04-10
Optionally limit new crate publishing by non-admin users.

maciekglowka:
The PR introduces a configuration option `new_crates_restricted` that would disallow non-admin users to publish new crates (without prior versions existing).

This is accompanied by the possibility to add empty crate placeholders (by admins only). These placeholders would be initialized with a 0.0.0 version and allow adding crate owners.
The existence of the placeholder would allow the owner-user to publish the first version of the crate (event with the restriction enabled).

This way we can add the admins more control over what crates are published and by whom. In other words, admins would have to pre-approve each crate and it's owner.

Currently those functionalities are available via the API only (as is also adding an owner to the crate). Once this is accepted it is also possible to add UI in the admin panel for that as a separate PR.

[Also I am sorry, but my rustfmt auto-formatted some otherwise untouched parts of the existing code. If there are are any specific formatting settings for kellnr I am happy to update the PR with them :]


maciekglowka:
@secana Hi, should we move this feature to issues for a discussion (as the new contributing guide prefers) or is it just waiting for a bit of free time (which I perfectly understand - so no pressure at all)?

Just asking to check the status, as we will probably rely on this feature and might have to use a fork (which is obv. not ideal ;) )

secana:
Hi. It's not needed to move it to a disussion. But I have some questions: What is the use-case of this? You wrote that you depend on it, but I cannot image in which context this is a needed freature. 
Second question: Do you have an idea for a more elegant solution to the "0.0.0 crate pushed by admin" approach? It feels a bit hacky to pre-approve crates this way. But to be honest, I don't have a better idea.

maciekglowka:
Hi, I think one use-case could be a repository for curated hardware related crates. One could imagine that only verified manufacturers would be able to publish crates for their own hardware (which would be marked e.g. in the crates name).

As for the second question, the only other solution I can think of is creating a separate table of reserved crates. Upon publishing of the first version it'd be checked whether there is such a crate-name reserved and the user has rights to it.
I am not sure if that's much more elegant, (kinda duplicating the data?). We could remove the entries from the table upon the first publish, but that could create problems on yanking...so I'd keep it. (unless yanking leaves a trace in the db anyway? I do not remember now).
If we removed the entries, I guess the table could be as simple as (crate_name | user_fk) - since the crate names are unique.

jauhien:
@secana , regarding the use-case: we are going to have a crates registry that is available for registered users, but we want to restrict the possibility to publish new crates, so user can publish only after approval.

jauhien:
Just for information: we are waiting with fixing merge conflict until there is conclusion on the way of the implementation.

secana:
Hi. Sorry for the late response, I was away for some time. I do not object to this feature, so feel free to update the PR. I'll release a new version as soon as it works. 

One additional question: Would namespaces be interesting for your use-case? It one of the most wanted features for crates.io and it seems that it could be interesting for you.

maciekglowka:
Thanks for the reply. I have rebased the PR (the 0.0.0 crate one) to the devel branch and did necessary updates.
As for the Namespaces, I don't immediately see a use case for us, but we'll talk it through internally (I might be unaware ;)

secana:
Thanks for the updated PR. I'll give the whole devel branch some more testing before the next kellnr release. Is there anything that you want to have in that release additionally or should we release "as is"?

Issue #675, opened 2025-04-11
Hardening data integrity on the DB level


maciekglowka:
Hi,

when I was working on another PR I have noticed that it is possible sometimes to put the DB into inconsistent state. You can e.g. add the crate owner twice via the API and get duplicate entries in the table. Then, when removing the owner, only one row will be deleted. Which is not ideal as, the user will still have owner privileges.

As it's IMO more of a data integrity issue, than a business logic, I thought this could be solved right at the DB. A simple migration adding a unique index seems to be enough:

```rust
        let owner_index = Index::create()
            .name("idx-owner")
            .table(OwnerIden::Table)
            .col(OwnerIden::CrateFk)
            .col(OwnerIden::UserFk)
            .unique()
            .to_owned();

        manager.create_index(owner_index).await?;
```

Similar indices could be applied for the `crate_index` table (on columns krate_fk and version) and perhaps `crate_user`. (perhaps some more, I do not know the code that well). In this cases the API endpoints actually do check for invalid data, but the database functions could as well be called from elsewhere in the code as well (if not now, then possibly in the future). So, at least in theory, it is possible to have two rows in the db for the same crate version - which I guess might also lead to some confusion. (I have also not tested what would happen if two similar requests came at the same moment)

Another issue would be perhaps to consider securing some DB calls under a transaction. Esp. the `add_crate` and `remove_crate` methods perform quite some number of db operations inside. If anything fails in the middle we'll end up in a partial data update. For example this is possible:

```rust
    test_db
        .db
        .test_add_crate(
            "mycrate",
            "admin",
            &Version::try_from("1.0.0").unwrap(),
            &Utc::now(),
        )
        .await
        .unwrap();
    test_db
        .db
        .test_add_crate(
            "mycrate",
            "user",
            &Version::try_from("2.0.0").unwrap(),
            &Utc::now(),
        )
        .await;
    // this is err due to an invalid user

    let max_version = test_db
        .db
        .get_max_version_from_name(&NormalizedName::from_unchecked("mycrate".to_string()))
        .await
        .unwrap();

    // /this fails as the max version got already bumped to 2.0.0
    assert_eq!("1.0.0", max_version.to_string());
}
```

Firstly the `krate` table got updated, and only then tables involving username are modified (which fails). Again, an invalid user won't be able to call the API endpoint as the auth would fail, but still there are a lot of calls so something can potentially go wrong here :) 

```rust
        self.add_owner_if_not_exists(owner, crate_id).await?;
[...]

PR #679, opened 2025-04-15
Use transaction for complex db inserts

maciekglowka:
Initial solution for the second part of #675 

Refactor of the `Database` type: internal methods moved to stand-alone functions, that take db connection as an argument. The functions are now generic over the `ConnectionTrait`, as this seems to be the convention that SeaOrm follows as well.
`Provider` implementation methods that make multiple changes to the DB have been wrapped in a transaction, to assure data consistency on failure.
Tests have been updated.


secana:
Could you please rebase the PR against the "devel" branch and format it with the default "cargo fmt". That should make the PR much smaller.

maciekglowka:
Hi again. I've rebased to devel and changed the PR target to the branch as well (as this seems to be the procedure now, according to the contributing guide).
Indeed I had `edition="2021" in my `rustfmt.toml` - hence the differences. But I've changed it to a blank default now and reformatted.

I am afraid github's differ doesn't handle well the total refactor of the `database.rs` file. The diff view is quite a mess :/ 
Anyway I've applied the differences between main and devel (mostly token hashing) - so it should be now fully compatible with the current devel branch. (initialy I worked on main). At least the tests seem to pass locally :)

PR #720, opened 2025-06-09
style: inline format args

nyurik:
```
cargo clippy --fix --allow-dirty --workspace --all-targets --all-features
cargo fmt
```

manually fixed `write!(f, "{}", &self.0)` into `write!(f, "{}", self.0)` -- this is faster (compiler is unable to optimize these, and results in about 6% perf cost)

manually fixed a few more cases when the same thing is done in tracing/logging


secana:
You're on fire! Merged!

PR #721, opened 2025-06-09
chore: lots of simple lint cleanup

nyurik:


crates/auth/src/token.rs
```diff
@@ -11,6 +11,7 @@ use std::sync::Arc;
 
 #[derive(Debug)]
 pub struct Token {
+    #[expect(clippy::struct_field_names)]
```
secana: I don't like this lint. Overlapping struct and field names are too common to have a lint for it and it adds a unnecessary line of code.

crates/db/src/database.rs
```diff
@@ -242,7 +242,7 @@ impl DbProvider for Database {
         let s = session::ActiveModel {
             token: Set(session_token.to_owned()),
             created: Set(created),
-            user_fk: Set(user.id as i64),
+            user_fk: Set(i64::from(user.id)),
```
secana: Not a fan of ``::from()` instead of `as`. I think `as` is easier to read and makes the intend of casting more clear.

crates/db/src/database.rs
```diff
@@ -748,7 +748,7 @@ impl DbProvider for Database {
     }
 
     async fn delete_auth_token(&self, id: i32) -> DbResult<()> {
-        auth_token::Entity::delete_by_id(id as i64)
+        auth_token::Entity::delete_by_id(i64::from(id))
```
secana: See comment above

crates/db/src/database.rs
```diff
@@ -967,7 +967,7 @@ impl DbProvider for Database {
             .columns(vec![CrateIden::OriginalName, CrateIden::TotalDownloads])
             .from(CrateIden::Table)
             .order_by(CrateIden::TotalDownloads, Order::Desc)
-            .limit(top as u64)
+            .limit(u64::from(top))
```
secana: See comment above

crates/db/src/database.rs
```diff
@@ -1378,7 +1378,12 @@ impl DbProvider for Database {
                     let mut ft = Vec::new();
                     for dep in ix {
                         ft.push(CrateRegistryDep::from_index(
-                            get_desc_for_crate_dep(&self.db_con, &dep.name, &dep.registry).await?,
+                            get_desc_for_crate_dep(
+                                &self.db_con,
+                                &dep.name,
+                                dep.registry.as_deref(),
```
secana, line 1376: `&` seems more concise than `as_deref()`

secana (changes requested):

crates/auth/src/token.rs
nyurik: sure, I can disable it crate-wide.

In this specific case I would suggest renaming it to `value` -- to make usage cleaner:

```rust
let token = Token::new(...);
if token.value == '...' {
  ...
}
```

Or it could be `.value()` with an accessor function. Otherwise it would be `token.token`.

P.S. this is the only case of `struct_field_names` in the entire code, so renaming might be easier?

crates/db/src/database.rs
nyurik: See https://rust-lang.github.io/rust-clippy/master/index.html?groups=pedantic#ref_option -- and especially see the video linked from that page

Disclaimer:  I wrote that lint

crates/db/src/database.rs
nyurik: ok, reverted.

Do note that `as` may cause some surprises, and Rust is all about eliminating those. In a way, `as` is like using `unwrap` on a `Result` because we assume many things will just work. Except that it is even worse - casting to smaller value drops some bits, which may produce incorrect result rather than panicking.  In any case, not that big of a deal, all's good :)

crates/auth/src/token.rs
secana: Yes, renaming it to `value` makes sense in that case.

nyurik:
should i submit `value` rename now or separate PR?

secana:
> should i submit `value` rename now or separate PR?

You can do it here. It's such a minimal and purely internal change that a new PR is overkill.

nyurik:
ok, rebased and renamed

PR #722, opened 2025-06-09
chore: cleanup use wildcards

nyurik:


secana:
Not happy with this one, as it make the code more verbose.

nyurik:
I just realized the verbose ones are test-like -- except that they are not excluded with `cfg(test)` for various reasons. Would this change make sense?  I agree wildcards in tests work well, but having wildcards in the non-test code is not that great and is not used frequently due to potential for conflicts in the updated dependencies.

secana:
Yes, I agree. In test code it is fine, in non-test code we should get rid of wildcard imports.

nyurik:
cool, i already did that change - hopefully it looks better.
Turns out this lint already behaves this way - it allows tests and public re-exports.  See https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports notes at the bottom.

Turns out it is also possible to use `allowed-wildcard-imports = [ "utils", "common" ]` pattern - where you can say which modules are "test-like" i guess?  I could try it out too

nyurik:
i tried it, and `allowed-wildcard-imports = ["super", "mockall"]` works, but I think `#[expect(...)]` is cleaner here, so seems this is good to go

PR #728, opened 2025-06-11
chore: optimize dependencies

nyurik:
Sort all dependencies alphabetically and remove some dependencies that are not needed.

```
cargo +nightly udeps --workspace --all-targets --all-features
```


secana:
> Sort all dependencies alphabetically and remove some dependencies that are not needed.
> 
> ```
> cargo +nightly udeps --workspace --all-targets --all-features
> ```

I had no idea that this exists. That's great!

PR #730, opened 2025-06-11
chore: fix `unused_async`

nyurik:
Fixes the [unused_async](https://rust-lang.github.io/rust-clippy/master/index.html#unused_async). Turns out there is a lot of complexity due to everything being async. So following the YAGNII principle, keeping things simpler is a good practice. Note that this means a few functions that actually do need to be async are marked as such with an `allow` attribute. This change also simplified a few unneeded `Result<T>` into `T`.


nyurik:
It seems the CI error is unrelated

secana (approved):

PR #736, opened 2025-06-16
[WIP] Implement separate authentication requirements for the UI vs crates API

### LINKED ISSUE #712: Separte crates API authentication from UI
Contrary to #93 there **is** a use-case for having the authentication on the API without also requiring it for the UI:

I am currently working on an internal kellnr deployment where we want it to be completely private. Only relying on kellnr's token auth implementation for the API, while letting our existing OAuth middleware handle access to the non-admin web UI.

This is mostly a QoL decision where for day-to-day read-only activities, like browsing the docs, we don't want to have to go trough the web form auth on top of our standard OAuth. Obviously for admin functionality, like token management, we'd still require kellner's auth.

We're using Traefik & Nomad. For those looking to do something similar the config looks something like this:

```nomad
tags = [
          "traefik.enable=true",

          "traefik.http.routers.kellnr-web.rule=Host(`kellnr.${DOMAIN}`)",
          "traefik.http.routers.kellnr-web.service=kellnr-web",
          "traefik.http.services.kellnr-web.loadbalancer.server.port=${NOMAD_HOST_PORT_http}",
          "traefik.http.routers.kellnr-web.middlewares=auth",

          "traefik.http.routers.kellnr-api.rule=Host(`kellnr.${DOMAIN}`) && PathPrefix(`/api`)",
          "traefik.http.routers.kellnr-api.service=kellnr-api",
          "traefik.http.services.kellnr-api.loadbalancer.server.port=${NOMAD_HOST_PORT_http}",
]
```

So by default, all of the traffic to the subdomain is hidden behind the `auth` middleware that takes care of OAuth. While any requests to `/api` path prefix have the `auth` middleware disabled.

## Ideal Solution

Have a way to configure kellnr such that only the API requires token auth, while non-admin web functionality does not. This enables the above setup which would alleviate (but not necessarily eliminate) the need for #695

Perhaps changing `auth_required` from `bool` to an `enum` like so:

```rust
enum AuthRequired {
    No, // Same as current `false`
    All, // Same as current `true`
    OnlyApi, // All `/api` requests require an auth token with appropriate permissions, but UI requests are treated as pre-authorized
}
```

If it helps, am happy to work on an implementation with some guidance on a preferred solution.

### Multiple Instances

A possible workaround I've considered is running two separate instances of kellner using the same data dir and db backend, but with different auth configration.

On top of being somewhat wasteful in terms of compute resources (running more instances than we actually need), it is unclear if such deployment is even supported.

The concern is about potential consistency issues across all possible combinations of local/s3 and sqlite/postgresql configurations.

### kellnr OAuth

On surface, #695 looks like it would completely address this.

But, in our case, it would still be sub-optimal. As we'd have to have separate OAauth configuration that could drift from our existing global configuration with undesirable consequences.

nCrazed:
Initial draft of the implementation.

I am not too happy with the `for_api()` and `for_ui()` functions, any suggestions for a cleaner interface are welcome

Resolves #712 

TODO
=====

- [ ] Finalize type names
- [ ] Update all docs
- [ ] Add (conditional?) deprecated marker to the boolean format
- [ ] Verify which routes should/should not be covered by the new config value

### DIFF
--- a/crates/auth/src/auth_req_token.rs
+++ b/crates/auth/src/auth_req_token.rs
@@ -13,8 +13,8 @@ pub async fn cargo_auth_when_required(
     request: Request,
     next: Next,
 ) -> Result<Response, StatusCode> {
-    if !state.settings.registry.auth_required {
-        // If auth_required is not true, pass through.
+    if !state.settings.registry.auth_required.for_api() {
+        // If crates API authentication is not required, pass through.
         return Ok(next.run(request).await);
     }
 
@@ -42,6 +42,7 @@ mod test {
     use db::mock::MockDb;
     use mockall::predicate::*;
     use settings::Settings;
+    use settings::registry::LegacyAuthRequiredWrapper;
     use std::sync::Arc;
     use tower::ServiceExt;
 
@@ -49,7 +50,7 @@ mod test {
     async fn no_auth_required() {
         let settings = Settings {
             registry: settings::Registry {
-                auth_required: false,
+                auth_required: LegacyAuthRequiredWrapper::Deprecated(false),
                 ..settings::Registry::default()
             },
             ..Settings::default()
@@ -67,7 +68,7 @@ mod test {
     async fn auth_required_but_not_provided() {
         let settings = Settings {
             registry: settings::Registry {
-                auth_required: true,
+                auth_required: LegacyAuthRequiredWrapper::Deprecated(true),
                 ..settings::Registry::default()
             },
             ..Settings::default()
@@ -85,7 +86,7 @@ mod test {
     async fn auth_required_but_wrong_token_provided() {
         let settings = Settings {
             registry: settings::Registry {
-                auth_required: true,
+                auth_required: LegacyAuthRequiredWrapper::Deprecated(true),
                 ..settings::Registry::default()
             },
             ..Settings::default()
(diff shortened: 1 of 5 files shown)


secana:
I'm fine with the `for_api` and `for_ui` functions. It's not sexy but it is simple and easy to understand. We can still think of something more sophisticated if the feature works as intended, if we want to.

nCrazed:
@secana routes defined in `kellnr_api_routes` are used by the UI frontend, right?

So realistically they should be excluded from the `OnlyApi` case (which should probably be renamed to something like `OnlyCratesApi`)? Otherwise we'd have a case where where an anonymous user can navigate to pages as expected, but have them partially broken due to failing API calls

secana:
Yes these routes are used by the frontend, as well as by cargo. That makes the distinction between UI frontend and cargo access difficult, but some of the functionality exists in both.

nCrazed:
Oh, I was operating under the assumption that cargo would only be using  `/api/v1/cratesio` is this not correct? 

If not, what other routes should be considered? Is this feature still viable?

secana:
`/api/v1/cratesio` are the routes for the crates.io proxy, while the `/api/v1/crates` (from kellnr_api) are routes for crates hosted on kellnr. Kellnr provides two distinct indicies to prevent duplicate names. E.g. a crate from kellnr named "foobar" vs. a crate with the same name from crates.io.

You can find all routes that the web ui uses in https://github.com/kellnr/kellnr/blob/main/ui/src/remote-routes.ts

I hope that helps.

PR #765, opened 2025-08-07
ui/nix update to support x86_64 platform build and updated node2nix.

giacomocariello:
I've recently tried to build kellnr on nix with platform linux_amd64, but received some error due to missing esbuild/linux-x64 npm cached package. Therefore, I've updated relevant nix files and package-patch.json to include support for x64 version.


secana:
thanke for the PR!

PR #773, opened 2025-08-22
fix(crates/auth): handle bearer token and allow config.json even thro…

FlorentinDUBOIS:
…ugh force authentication


secana:
Hi! Thanks for putting in the effort. Could you please re-base the PR to go against the `devel' branch instead of `main`? The `devel` contains the code for the next Kellnr version. As the changes are small, I'm happy to include them in the next release.

FlorentinDUBOIS:
I have rebased the branch your `devel` and change the target branch of the pull request.
I am glad that you can include them in the next release :)

secana:
Thanks for rebasing. Can you check the broken test?

crates/auth/src/auth_req_token.rs
```diff
@@ -12,19 +13,31 @@ pub async fn cargo_auth_when_required(
     State(state): State<appstate::AppStateData>,
     request: Request,
     next: Next,
-) -> Result<Response, StatusCode> {
-    if !state.settings.registry.auth_required {
+) -> Response {
+    // The path /config.json is always accessible even if auth_required is true
+    // it is used for cargo and others registry as nexus to know if we should provide a token or not
```
BradG13542: Just curious but is this accurate / does this PR still work without these changed lines?

The Cargo reference [here](https://doc.rust-lang.org/cargo/reference/registry-index.html#sparse-authentication) mentions how Cargo should retry getting `config.json` with auth if the registry initially returns a 401.

Great changes overall though 👍

secana: Great catch! `... || request.uri().path() == "/config.json"` is something we must not do, as we leak the `config.json` to unauthorized users. @FlorentinDUBOIS is this required for you?

FlorentinDUBOIS: I think it is not required, I had to expose it for the nexus3 and artifactory, but I think it is because the authentication part is not yet implemented, I will remove it

FlorentinDUBOIS: I have updated the pull request with the modification, rebased it on the devel branch and change the title (of the commit as well) to better reflect changes.

FlorentinDUBOIS:
> Thanks for rebasing. Can you check the broken test?

I have fix the import issue, should be fine by now :) 

Thank you for the review, I hope I am not too late for the release :)

PR #800, opened 2025-09-12
Feat: Add cratesio indices

fellhorn:
The crates-io tables were missing some indices, which had a negative impact on DB performance due to unnecessary table scans.

In particular this adds indices for:

`cratesio_index.crates_io_fk`
`cratesio_meta.(crates_io_fk && version)`
`cratesio_crate.name`


crates/db/entity/src/auth_token.rs
```diff
@@ -1,4 +1,4 @@
-//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.4
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.16
```
fellhorn, line 1: Let me know if I should generate them with the old sea-orm version instead

crates/db/entity/src/auth_token.rs
```diff
@@ -9,7 +9,7 @@ pub struct Model {
     pub id: i64,
     #[sea_orm(column_type = "Text")]
     pub name: String,
-    #[sea_orm(column_type = "Text")]
+    #[sea_orm(column_type = "Text", unique)]
```
fellhorn, line 12: This was already part of a previous migration. Looks like the entity files have not been updated?

Issue #153, opened 2023-12-04
Publish webhook or something similar


jfro:
Was thinking it'd be nice to have kellnr hit a webhook or run a script upon successful publish, notifying of newly published crate/version to something like Slack.

secana:
That's a cool idea and doesn't seem too much work. I'll look into it.

secana:
To support a webhook we need:

- [ ] POST endpoint `/api/v1/webhook` to register a hook with `register` body
- [ ] DELETE endpoint `/api/v1/webhook?id=x` to remove a webhook
- [ ] GET endpoint `/api/v1/webhook` to list all registered hooks or `/api/v1/webhook?id=x` to get info about a specific hook
- [ ] All API calls have to be authenticated and only admin accounts are allowed
- [ ] A library that contains the events to be used to serialize and deserialize events. Public on crates.io
- [ ] Store all registered events in the database to be persistent
- [ ] Add UI to see registered hooks
- [ ] Add UI to register hooks
- [ ] Add UI to delete hooks

The `register` event needs to contain all relevant data for kellnr, to be able to call the client in case, a event occurs. 

```json
// register event
{
   "url": "the url that gets a POST event",
   "name": "optional name for the hook",
   "events": [
       "crate_add",
       "crate_update",
       "crate_delete"
    ]
}
```

If the `register` event was send to kellnr, it responses with a `HTTP OK` and an `ID` that the client can use to identify the hook.

In case of an event, kellnr sends a `HTTP POST` to the registered urls.
To distinguish the event types, append a parameter to the url, e.g. `https://url?event=crate_add`

```json
// event
{
   "id": "x",
    "name": "optional name",
    "event": "crate_add".
     "data": {
     // relevant data for the event
     }
}
```

secana:
The webhook API can be a first step to support a separate builder container, to keep the kellnr image small. If a new crate/version is uploaded, call a webhook to the builder container. #276

secana:
Adhere to the https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md where possible

PR #814, opened 2025-09-24
Initial Webhook API

maciekglowka:
## Basic webhook support based on:
https://github.com/kellnr/kellnr/issues/153

This is a WIP PR meant for a further discussion.

### What has been done:

- initial API that allows to register, view, delete webhooks
- working webhooks for `crate_add`, `crate_update`, `crate_yank`, `crate_unyank` events
- a test endpoint - meant to validate webhook connectivity
- retry mechanism when send fails

### What has not been done:
- tests (due to draft status and possible changes in the code)
- documentation (same as above)
- UI (beyond the scope of this PR for me currently)

### API call examples:

Register
```
curl kellnr_url/api/v1/webhook -v -X POST -H "Authorization: Bearer <TOKEN>" -H "Content-Type: application/json" -d '{"action": "crate_unyank", "callback_url": "http://192.168.100.140:8000", "name": "My unyank webhook"}'
```

Test
```
curl kellnr_url/api/v1/webhook/170c3a3c-75a8-4597-864e-bde03e313027/test -v -X GET -H "Authorization: Bearer <TOKEN>"
```

Get all
```
curl kellnr_url/api/v1/webhook -v -X GET -H "Authorization: Bearer <TOKEN>"
```

Get single
```
curl kellnr_url/api/v1/webhook/170c3a3c-75a8-4597-864e-bde03e313027 -v -X GET -H "Authorization: Bearer <TOKEN>"
```

Remove
```
curl kellnr_url/api/v1/webhook/170c3a3c-75a8-4597-864e-bde03e313027 -v -X DELETE -H "Authorization: Bearer <TOKEN>"
```

### Sample callback payload

```json
{
  "data": { "crate_name": "test_crate", "crate_version": "0.1.2" },
  "timestamp": "2025-09-24T15:19:04.842575427Z",
  "type": "crate_yank"
}
```

### Remarks

- each action / event requires a separate webhook registration (unlike described in the issue) - however same url can be used many times; I believe in the future there might be other types of events (crate unrelated, like user management) so, the target service might not be interested in receiving many unwanted notifications
- there is no url parameter added (`?crate_add`) to the callback request; as the action is described in the `type` field of the payload I believe this is redundant
- since the webhook ids are shared externally I've felt is is more natural / safe to use uuids rather than ints; for consistency sake webhook_queue ids are also uuid (despite being only used internally)
- authorization: an admin token is used - I am not sure if that is compatible with a future UI attempt; the UI endpoints seem to be using a `MaybeUser` struct to verify permissions - perhaps this should be used? (however that would mean sending admins username and password in the API calls?)

Naturally, all of the above are just my proposal - meant for a discussion ;)

### DIFF
--- a/crates/db/src/provider.rs
+++ b/crates/db/src/provider.rs
@@ -170,7 +170,7 @@ pub trait DbProvider: Send + Sync {
         event: WebhookEvent,
         payload: serde_json::Value,
     ) -> DbResult<()>;
-    /// Extracts webhook queue entries with `next_attempt` at or earlier than proviced timestamp.
+    /// Extracts webhook queue entries with `next_attempt` at or earlier than provided timestamp.
     async fn get_pending_webhook_queue_entries(
         &self,
         timestamp: DateTime<Utc>,
@@ -474,34 +474,34 @@ pub mod mock {
                         unimplemented!()
             }
             async fn get_group_users(&self, group_name: &str) -> DbResult<Vec<User>> {
-                uninplemented!()
+                unimplemented!()
             }
             async fn is_group_user(&self, group_name: &str, user: &str) -> DbResult<bool> {
-                uninplemented!()
+                unimplemented!()
             }
 
             async fn add_crate_group(&self, crate_name: &NormalizedName, group: &str) -> DbResult<()>{
-                uninplemented!()
+                unimplemented!()
             }
             async fn delete_crate_group(&self, crate_name: &NormalizedName, group: &str) -> DbResult<()>{
-                uninplemented!()
+                unimplemented!()
             }
             async fn get_crate_groups(&self, crate_name: &NormalizedName) -> DbResult<Vec<Group>>{
-                uninplemented!()
+                unimplemented!()
             }
             async fn is_crate_group(&self, crate_name: &NormalizedName, group: &str) -> DbResult<bool>{
-                uninplemented!()
+                unimplemented!()
             }
 
             async fn is_crate_group_user(&self, crate_name: &NormalizedName, user: &str) -> DbResult<bool>{
-                uninplemented!()
+                unimplemented!()
             }
 
             async fn register_webhook(
                 &self,
                 webhook: Webhook
             ) -> DbResult<String> {
-                uninplemented!()
+                unimplemented!()
             }
             async fn delete_webhook(&self, id: &str) -> DbResult<()> {
                 unimplemented!()
--- a/crates/kellnr/src/routes/webhook_routes.rs
+++ b/crates/kellnr/src/routes/webhook_routes.rs
@@ -4,7 +4,7 @@ use axum::{
     Router,
 };
 
-/// Creates the user routes
+/// Creates the webhook routes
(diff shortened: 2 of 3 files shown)


secana:
Hi! I really like what you did there. For the user: I think requiring an admin user is necessary for that feature. Normal users should not be able to interact with webhooks. For a future UI, only the admin users should be able to see/edit the hooks. As there is already a differentiation for admin and normal user settings, this should work out fine. 

As you are already on the feature and I'm fine with your proposal, how would you like to proceed?

maciekglowka:
Hi! Thanks and sorry for a bit late reply (life got me a bit more last week ;)

If you have no serious / general remarks than I'd add some tests now. Probably when I'll look at the code again I might want to change a thing or two - as it's usually the case.
If the tests will work and pass, I'd head to the documentation.

secana:
Take you time, we all have other things to do. Ping me, when you need something!

maciekglowka:
Hi! I've finally got some time to work on this. (sorry took so long!)

I have mostly included some tests. Also renamed webhook `actions` to `events`, as it makes more sense I think.

There is perhaps one controversial thing, The Github standards (referred in the original issue) suggest an event field to be named `type` - which obviously is a reserved keyword in Rust. So currently I have named the field `event` everywhere internally. The public API uses `type` though - as the standard suggests. Which might lead to some confusion when working on code? (for incoming calls this is handled via serde alias, for outgoing it's more straightforward as the payload is constructed manually anyway).

 See https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#payload

secana:
Hi @maciekglowka! Sorry for the very late response, I was on a longer vacation. I've reviewed the code and it looks good to me. Copilot catched some issues, mostly typos. Would you be so kind and fix them? If that's done, I'll merge. Ofcourse, take your time as X-Mas and New Year are more important than fixing some nitpicks.

maciekglowka:
> Hi @maciekglowka! Sorry for the very late response, I was on a longer vacation. I've reviewed the code and it looks good to me. Copilot catched some issues, mostly typos. Would you be so kind and fix them? If that's done, I'll merge. Ofcourse, take your time as X-Mas and New Year are more important than fixing some nitpicks.

Thanks for the review. I've fixed the typos - didn't take too much time ;)
I've only left out a remark about extracting a ternary to a separate function. IMO it's an overkill and won't make it more readable in the future (it's not used multiple times) - however if you'd agree with the copilot I'm happy to change it (shouldn't take more than a minute ;)

secana:
@maciekglowka Thanks for fixing the typos. I agree that the rest is not necessary. I'll merge the PR and the feature will be in the next release. Thank you very much for the effort and have nice x-mas and happy new year!

maciekglowka:
> @maciekglowka Thanks for fixing the typos. I agree that the rest is not necessary. I'll merge the PR and the feature will be in the next release. Thank you very much for the effort and have nice x-mas and happy new year!

Thanks. Happy new year to you too ;)

PR #815, opened 2025-09-24
Add support for running kellnr on a URL subpath

hw0lff:


secana:
Hi! Cool feature, thanks for the PR. Would you mind adding corresponding documentation to https://github.com/kellnr/website such that the feature is discoverable on https://kellnr.io/documentation?

hw0lff:
> Would you mind adding corresponding documentation to https://github.com/kellnr/website such that the feature is discoverable on https://kellnr.io/documentation?

Will do, once I got kellnr with these changes integrated into our infra.

While building a docker container for our infra, I noticed an issue with the docker health check.
`HEALTHCHECK --interval=5m --timeout=3s CMD curl -f http://localhost:8000/api/v1/health || exit 1` will not work when changing the path.
Right now, I'm not sure where it needs fixing so both proxied and non-proxied setups can be supported with a working health check.
Any ideas?

secana:
Making the health check URL settable by the user could be an option. Very simple example:

```bash
FROM ubuntu:24.04

ENV HEALTHCHECK_URL=http://localhost:8000/api/v1/health

RUN apt-get update && apt-get install -y \
  curl

HEALTHCHECK --interval=10s --timeout=3s CMD curl -f ${HEALTHCHECK_URL} || exit 1

CMD ["sleep", "infinity"]
```

That should a user allow to set the URL to whatever needed with `docker run -e HEALTHCHECK_URL=http://foo.bar ...`. Would that work for you?

Not sure if it makes sense to make the whole URL settable. Maybe the optional path is enough, as if the whole URL can be changed, the container can be forced to request any resource on the internet.

hw0lff:
[commit 1dddb34](https://github.com/kellnr/kellnr/pull/815/commits/1dddb349571ccade1a859bd065cb8ae12dd1b582) removes the serving below the subpath. Let's assume the following config:
```toml
[local]
ip = "0.0.0.0"
port = 8000

[origin]
hostname = "srv.example.com"
port = 443
protocol = "https"
path = "/namespace/kellnring/"
```

Will let kellnr bind to `0.0.0.0:8000` and therefore serve `http://localhost:8000/api/v1/crates/config.json`.
`curl http://localhost:8000/api/v1/crates/config.json` reveals this json:

```json
{
  "dl": "https://srv.example.com:443/namespace/kellnring/api/v1/crates/dl",
  "api": "https://srv.example.com:443/namespace/kellnring",
  "auth-required": true
}
```

Now configuring cargo to use this index `sparse+http://localhost:8000/api/v1/crates/` will not work.
imo it doesn't need to because if you concern yourself with URL namespacing, you're likely running a reverse proxy already, allowing you to rewrite the URL if needed.
Have I overlooked something in my assumptions?

This let's us keep the health check in the container the same.

secana:
You are right. This should already work without the need for any code changes, expect the `path` in the origin. I don't currently have a setup to test it tho.

crates/index/src/config_json.rs
```diff
@@ -19,17 +19,22 @@ impl ConfigJson {
         protocol: Protocol,
         api_address: &str,
         api_port: u16,
+        api_url_path: &str,
```
secana: Would be nice if this is an `Option`. This would mean that we can hand around a `None` instead of `""`, which is more expressive.

secana (changes requested):

secana:
@hw0lff As far as I can see, only the change to the `Option` is missing to get this merged.

hw0lff:
I didn't add the documentation to the kellnr/website repo yet…

secana:
Thanks for the change. I'll wait for the docs an then merge both.

PR #825, opened 2025-10-01
Unify crypto

hw0lff:
This PR replaces self-made password hashing with sha256 with libsodiums pwhash (which uses argon2id).
According to https://latacora.github.io/blog/2024/07/29/crypto-right-answers-pq/#password-handling sha256 should not be used for password hashing.
sha256 is fine for file hashing.

I also collected all crypto related stuff into the `common::crypto` module, like token generation, random string generation.
The `crypto` module is supposed to be an opaque API for doing cryptography stuff, hiding inner implementations and should be used from now on instead of copying functions from other modules or importing `rand` directly.

The problem that arises when switching password hashing algorithms is that you cannot update all password hashes immediately. 
I tried to design the `crypto` module in a way that makes updating to newer crypto primitives easier to understand.
Once everything is migrated the `crypto::update` module will be in a state that it only contains obsolete functions and should be fairly easy to remove completely from the code base.

Passwords will be rehashed with argon2id once a user with an old pwhash logs in.
`generate_salt()` should be phased out because libsodium handles this for us.

unsolved problems/ breaking changes:
- all tokens will be invalid
- the `salt` field of the `User` type is obsolete and needs to be removed at a later migration
- I had to downgrade the `rand` module due to incompatibilites between versions


secana:
While I admit that SHA-256 is not the best fit for a password hash, as it has no input-of-work and is thus fast to hash, it is still secure. The proposed change would not increase the security for the user but pull in a crate (`alkali`) that has very few users and is not frequently updated, which is not something I expect from a crypto library.

While I recognize that you put a lot of effort into the PR and the idea with transparent updates of the algorithm is cool, I will deny this PR. I don't see the value vs. the added complexity. Sorry for that.

hw0lff:
Thank you for the feedback.
I mainly want to unify all crypto related stuff into a single place/module/crate.
The migration and switching to argon2id was just added on top.
I understand if this switch is not desired.

Still, I would like to consolidate all crypto stuff into a module.
So how about I reduce the changes to only that?
No migration stuff and no alkali.

The crypto module api would look like this:
```rs
mod crypto {
  pub fn store_password(password: &str, salt: &str) -> String;
  pub fn verify_password(password: &str, salt: &str, hash: &str) -> bool;
  pub fn generate_rand_string(length: usize) -> String;
  pub fn generate_salt() -> String;
  pub fn generate_token() -> String;
  pub fn store_token(token: &str) -> String;
  pub fn verify_token(token: &str, hash: &str) -> bool;
  pub fn hash_file_sha256(data: &[u8]) -> String;
}
```

It's important imo to communicate intend with the API.
So that just by reading you wouldn't use `hash_file_sha256` to hash a password because you have `store_password` for that.

Basically everything stays the same but the functions get deduplicated and collected into a single module, namely `common::crypto`.
Also the sha and rand crates are only direct dependencies for the `common` crate.

Would this be reduced changeset be alright?
It wouldn't leak implementation details of crypto stuff into the application, except for the salt.
It would also allow to implement constant-time comparison of hashes at a later time without any changes in the application code.

Or would you maybe rather have the crypo module be a CryptoProvider trait?
Do you have suggestion or are there any other changes that would be needed?

secana:
Hi! That sounds like a good idea and a PR with a crypto module is very welcome!

PR #843, opened 2025-10-27
Kellnr devcontainer

sfb103:
Adds files that support building and running a kellnr devcontainer, for those that prefer that dev env.


secana:
Hi! I do not use devcontainers, but nix instead. That's a lot of code to maintain and keep up-to-date such that it always works. As my expierience with devontainers is limited, is this something to be worried about? And what is the advantage over running `nix develop` and having all the needed dependencies installed while still having all your local development tools available?

sfb103:
Just noticed this... sorry for the delayed response... I don't have a lot of experience with nix, but def will give it a try.

My quick response is that nix seems like a great system for managing toolchains, but devcontainers can provide a more complete "development" env.  Best example of the diff is how devcontainers naturally integrates with VSCode (and forks like Cursor) and can even include specific VSCode extensions.  I've found devcontainers really useful for bouncing between projects which I do often.

That said, I'm excited to see first hand how nix compares and will give it a try.  Let's def hold off on this for now and after I've tried nix I'll let you know if it make sense to do add a devcontainer here or if it's overkill.

PR #873, opened 2025-12-10
feat: remove default S3 access/secret key

sfb103:
This allows the AmazonS3 crate to obtain credentials via a service account (IAM role) rather than secret key. The approach simply removes the definition for the secret key id and value from the default configuration.  With this absent, as long as no value is passed for this via the environment, these will not be provided to the s3 builder. Building the s3 client without these enables the crate to use a service account for obtaining credentials.

### DIFF
--- a/crates/storage/src/s3_storage.rs
+++ b/crates/storage/src/s3_storage.rs
@@ -65,7 +65,7 @@ impl TryFrom<(&str, &Settings)> for S3Storage {
     type Error = StorageError;
 
     fn try_from((bucket, settings): (&str, &Settings)) -> Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


nyurik:
@secana this seems ok, and should solve the long standing issue i was seeing. If no objection, i will merge and release

nyurik:
ah, darn, i forgot - this PR should be made against the `devel` branch rather than `main`

secana:
Hi! LGTM, you can merge into `main`. I'll rewrite the contribution guideline in the next days to use a simpler workflow with `main` and `feature` branches without `devel` inbetween.

@nyurik feel free to merge and release a new version. Let me know if you have questions or something does not work.

PR #911, opened 2025-12-31
test: improve smoke test speed

secana:
Use docker caching to improve the runtime of the smoke test. Use `debug` build instead of `release` in the smoke test to speed up the compilation and linking time.


secana:
While it's possible to speed the docker build up with caching, the added complexity it not worth the win.

PR #912, opened 2025-12-31
chore: use `license` field and other Cargo.toml cleanups

nyurik:
* The `license` field is cleaner and machine-readable (SPDX), whereas `license-file` should only be used for custom licenses.
* Added `authors` field to all cargo.toml to use the workspace value
* Switched to Workspace v3 resolver
* It might be better to use the usual Rust's `"MIT OR Apache-2.0"` value -- Apache gives more safety to the users, but not a big deal for an app.
* Many Cargo.toml blocks were sorted alphabetically, keeping workspace-shared values separate from the non-shared ones
* Remove explicit `[lib]` section - it is not needed, esp the name, because Cargo.toml automatically computes it based on available files.

### DIFF
--- a/crates/db/db_testcontainer/Cargo.toml
+++ b/crates/db/db_testcontainer/Cargo.toml
@@ -1,7 +1,9 @@
 [package]
 name = "db_testcontainer"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
 version.workspace = true
-edition = "2024"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


secana (approved):
I'm fine with switching to `MIT OR Apache-2.0`. You want to do it in this PR?

nyurik:
I think it would be better if you do it separately as the project owner.

PR #922, opened 2026-01-03
chore: format code and sort imports across multiple files

nyurik:
Add a new `just fmt` recipe. If nightly is installed, it will normalize the use statements: sorting, grouping by standard vs 3rd party vs internal code.  Without nightly, it simply runs `cargo fmt`.


secana (approved):
lgtm. Ready to merge.

Issue #928, opened 2026-01-06
Multiple instances cause intermittent 401 auth errors


nyurik:
We just rolled out a Kellnr service with 3 pods and postgreSQL/S3 backend, and the restricted APIs like `list_tokens` and `list_users`  fail with the empty `401 Unauthorized` response exactly 2/3 of the time - meaning that only the auth token issues from one of the pods is working, while two other pods fail.  Does Kellnr handle multi-host secret sharing, i.e. via postgres?

If token cannot be shared among pods, one option could be to set up multiple read-only pods plus a single read-write pod, and route ingress traffic to rw service in case of the request having `kellnr_session_id` cookie or the authentication headers. This would help with the common read-only cases of downloading existing crates (assuming no  authentication is needed for readonly access).

Lastly, the helm chart may need to enable `session affinity by client IP` (kinda tricky in some cases)

@vimoudy

secana:
Hi! kellnr was never tested running with more than one instance. May I ask that the use-case is? Having more than one instance will make cache-misses more likely. But if there is a good reason for having mutliple instances, I'll have a look why it fails.

nyurik:
resilience - if a large org depends on kellnr as part of the CI, we wouldn't want to break tons of CI processes just because kellnr happens to be updating or migrating to a new node cluster as part of routine k8s maintenance.

secana:
While kellnr stores the session token in the PostgreSQL, this won't help much as kellnr uses cryptographically signed cookies with a random key. The key is generated randomly when kellnr starts. If you have three instances of kellnr, each of them has another signing key. That's why the cookie cannot be validated by another instance, except the one that created the cookie.

Can you give it a try with setting the session affinity? If that works, that would be the easy solution.

If not, we need to think about changing the code. As I would not like to lower the security by hardcoding a deterministic key, the only option left is syncing the key over the DB, too. But that means that we need to think about key lifecycle management and add a bunch of database stuff, which is always a lot of work.

nyurik:
I think a simple solution would be to use a REDIS pod, or even a cluster with the [sentinel for fail-over](https://www.baeldung.com/ops/redis-sentinel-kubernetes-high-availability). No storage needed (ephemeral), resiliant MOST of the time, and very small memory footprint. If enabled, the security token would go to the shared redis store rather than local.

secana:
[...]

PR #951, opened 2026-01-15
feat: Configurable cookie signing key #928

secana:
Allow setting the UI authentication cookie signing key, to support multiple kellnr instances share the same login.

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -43,6 +43,8 @@ auth_required = false
 required_crate_fields = []
 # Disallows new crate publishing by the users.
 new_crates_restricted = false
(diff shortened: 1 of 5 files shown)


secana:
@nyurik I implemented a simple setting to add a shared cookie signing key. I tested it on my Kubernetes cluster with three instances and a Kubernetes secret that is shared by all pods, and it works fine.

The current implementation is very simple and does not refresh the key automatically, as this would be a lot more complicated and break all current sessions. The security win through that is so low that it’s not worth the hassle IMHO. If there is a need for a more sophisticated solution in the future, we can build on the simple implementation.

Would this PR solve your multi-replica issue?

PR #953, opened 2026-01-15
feat(auth/tokens): implement token cache, to relieve some pressure on db and in high throughput scenarios

perfectra1n:
Hi all,

Thanks again for Kellnr, it's been seriously awesome. For some local apps that I have and use self-hosted CI/CD for, I was getting rate limited by Cargo and this has been a life saver.

I had some strange errors though, when multiple jobs would run. Under high throughput scenarios (e.g., many concurrent crate downloads), the database connection pool can become exhausted. When this happens, `get_user_from_token()` fails, and all database errors are mapped to `403 FORBIDDEN`:

```rust
// Previous behavior in crates/auth/src/token.rs
Err(_) => Err(StatusCode::FORBIDDEN),
```

This caused legitimate, authenticated requests to receive false 403 errors during periods of high load, even though the tokens are valid. I experience intermittent authentication failures like:

```
error: failed to download `some-crate v1.0.0`
Caused by:
  failed to verify the checksum of `some-crate v1.0.0`
```

The token validation flow queries the database on every single request:

```
Request → Extract Token → Query DB → Validate → Response
                            ↑
                     Connection pool exhausted
                     under high load = 403 error
```

Due to this code in `crates/auth/src/token.rs`:
```rust
// Original - direct DB call on every request
let user = db
    .get_user_from_token(token)
    .await
    .map_err(|_| StatusCode::FORBIDDEN)?;
```

and this code in `crates/db/src/database.rs`:
```rust
async fn get_user_from_token(&self, token: &str) -> DbResult<User> {
    let token = hash_token(token);

    let u = user::Entity::find()
        .join(JoinType::InnerJoin, user::Relation::AuthToken.def())
        .filter(Expr::col((AuthTokenIden::Table, AuthTokenIden::Token)).eq(token))
        .one(&self.db_con)  // Direct SQL query every time
        .await?
        .ok_or(DbError::TokenNotFound)?;
    // ...
}
```

So I took a shot at implementing in-memory token caching with retry logic for transient database errors, while maintaining security through complete cache invalidation on token/user changes.

```
Request → Extract Token → Check Cache → Cache Hit? → Return cached data
                              ↓ (miss)
                         Query DB (with retries)
                              ↓
                         Cache result → Response
```

#### 1. Cache Location: `kellnr-common` crate

The `TokenCacheManager` lives in `kellnr-common` rather than `kellnr-auth` to avoid circular dependencies. The `kellnr-auth` depends on `kellnr-appstate` (for `AppStateData`), `kellnr-appstate` needs access to `TokenCacheManager`, and placing it in `kellnr-common` breaks this cycle. E.g.:

```rust
// crates/common/src/token_cache.rs
pub struct TokenCacheManager {
    cache: Option<Cache<String, CachedTokenData>>,
}

pub struct CachedTokenData {
    pub user: String,
    pub is_admin: bool,
    pub is_read_only: bool,
}
```

#### 2. Complete Cache Invalidation Strategy

Rather than tracking individual token invalidations (which is complex and error-prone), I used complete cache invalidation whenever any token-related change occurs (like tokens being added/deleted):

```rust
// Example: when a token is deleted
pub async fn delete_token(
    State(db): DbState,
    State(cache): TokenCacheState,  // Added
    // ...
) -> Result<(), RouteError> {
    db.delete_auth_token(&user, &data.name).await?;
    cache.invalidate_all().await;  // Invalidate entire cache
    Ok(())
}
```

Why complete invalidation?

Keeps the implementation simpler with fewer edge cases, and token operations are infrequent compared to reads. The 30-minute TTL provides natural cache refresh, and helps eliminate the risk of stale permissions persisting

Endpoints that trigger cache invalidation:
- `add_token()` - New token added
- `delete_token()` - Token removed
- `delete()` (user) - User deleted (all their tokens invalid)
- `read_only()` - User permissions changed
- `add()` (user) - New user added (admin status may differ)

#### 3. Retry Logic for Transient Errors

Not all database errors trigger retries. I distinguished between them via:

| Error Type | Retry? | Rationale |
|------------|--------|-----------|
| Connection timeout | Yes | Transient, may succeed on retry |
| Pool exhaustion | Yes | Transient, pool may free up |
| `TokenNotFound` | No | Definitive - token doesn't exist |
| `UserNotFound` | No | Definitive - user doesn't exist |

```rust
async fn get_user_with_retry(
    db: &Arc<dyn DbProvider>,
    token: &str,
    max_retries: u32,
    delay_ms: u64,
) -> Result<kellnr_db::User, DbError> {
    let mut last_error = None;

    for attempt in 0..=max_retries {
        match db.get_user_from_token(token).await {
            Ok(user) => return Ok(user),
            Err(e) => {
                // Don't retry on definitive "not found" errors
                if matches!(e, DbError::TokenNotFound | DbError::UserNotFound(_)) {
                    return Err(e);
                }
                // Retry on transient errors
                if attempt < max_retries {
                    tokio::time::sleep(Duration::from_millis(delay_ms)).await;
                }
                last_error = Some(e);
            }
        }
    }

    Err(last_error.unwrap())
}
```

#### 4. Security: No 503 Exposure

Similarly to what was already in the codebase, I chose to intentionally continue returning `403 FORBIDDEN` rather than `503 SERVICE UNAVAILABLE` to avoid information leakage.

#### 5. Disabled Cache Option

The cache can be disabled via configuration for debugging or environments where caching is undesirable (e.g. `KELLNR_REGISTRY__TOKEN_CACHE_ENABLED`):

```rust
impl TokenCacheManager {
    pub fn new(enabled: bool, ttl_seconds: u64, max_capacity: u64) -> Self {
        let cache = if enabled {
            Some(Cache::builder()
                .time_to_live(Duration::from_secs(ttl_seconds))
                .max_capacity(max_capacity)
                .build())
        } else {
            None  // All operations become no-ops
        };
        Self { cache }
    }
}
```

## Configuration

I added the new settings in `Registry`:

| Setting | Default | Description |
|---------|---------|-------------|
| `token_cache_enabled` | `true` | Enable/disable token caching |
| `token_cache_ttl_seconds` | `1800` | Cache TTL (30 minutes) |
| `token_cache_max_capacity` | `10000` | Maximum cached tokens |
| `token_db_retry_count` | `3` | Number of retry attempts |
| `token_db_retry_delay_ms` | `100` | Delay between retries |

## Test Coverage

I added 22 tests across three areas to make sure this all works correctly.

The token cache itself (crates/common/src/token_cache.rs) has basic unit tests covering the happy path (insert/get), disabled mode, cache misses, and the invalidate_all behavior.

For the retry logic (crates/auth/src/token.rs), I wanted to make sure we only retry on transient errors. So there are tests verifying that connection timeouts trigger retries while TokenNotFound and UserNotFound errors fail immediately without wasting time on retries. There's also coverage for edge cases like zero retries configured and first-attempt successes.

The cache integration tests (also in token.rs) verify the full flow: cache hits skip the DB entirely, cache misses populate the cache for next time, and invalid tokens don't get cached (so we don't accidentally cache a "not found" result).

Finally, the cache invalidation tests (crates/web-ui/src/user.rs) confirm that the cache gets cleared when tokens or users are added/deleted, and when permissions change. There's also a negative test making sure we don't invalidate the cache if the underlying DB operation fails—no point clearing good cached data if the change didn't actually happen.

## Performance

Cache hits are essentially free (~0ms in-memory lookup). Cache misses have the same latency as before plus a tiny overhead to store the result. Under load, the DB sees significantly less pressure since repeated token validations hit the cache instead of running queries. Memory-wise, we're looking at roughly 1MB max (about 100 bytes per token × 10,000 max capacity).

## Etc.

Hopefully you find these changes helpful, let me know if you would like something changed or done differently.

### DIFF
--- a/ui/src/components/StartupConfig.vue
+++ b/ui/src/components/StartupConfig.vue
@@ -177,6 +177,91 @@
                     </div>
                   </td>
                 </tr>
+                <tr>
+                  <td class="text-subtitle-2">Token Cache Enabled</td>
+                  <td>
+                    <v-chip :color="settings.registry.token_cache_enabled ? 'success' : 'grey'" size="small"
+                      text-color="white">
+                      {{ settings.registry.token_cache_enabled ? 'true' : 'false' }}
+                    </v-chip>
+                  </td>
+                  <td>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="deep-purple-darken-1" text-color="white"
+                        class="config-chip">TOML</v-chip>
+                      <span class="config-value">registry.token_cache_enabled</span>
+                    </div>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="teal-darken-1" text-color="white"
+                        class="config-chip">ENV</v-chip>
+                      <span class="config-value">KELLNR_REGISTRY__TOKEN_CACHE_ENABLED</span>
+                    </div>
+                  </td>
+                </tr>
+                <tr>
+                  <td class="text-subtitle-2">Token Cache TTL (seconds)</td>
+                  <td>{{ formatValue(settings.registry.token_cache_ttl_seconds) }}</td>
+                  <td>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="deep-purple-darken-1" text-color="white"
+                        class="config-chip">TOML</v-chip>
+                      <span class="config-value">registry.token_cache_ttl_seconds</span>
+                    </div>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="teal-darken-1" text-color="white"
+                        class="config-chip">ENV</v-chip>
+                      <span class="config-value">KELLNR_REGISTRY__TOKEN_CACHE_TTL_SECONDS</span>
+                    </div>
+                  </td>
+                </tr>
+                <tr>
+                  <td class="text-subtitle-2">Token Cache Max Capacity</td>
+                  <td>{{ formatValue(settings.registry.token_cache_max_capacity) }}</td>
+                  <td>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="deep-purple-darken-1" text-color="white"
+                        class="config-chip">TOML</v-chip>
+                      <span class="config-value">registry.token_cache_max_capacity</span>
+                    </div>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="teal-darken-1" text-color="white"
+                        class="config-chip">ENV</v-chip>
+                      <span class="config-value">KELLNR_REGISTRY__TOKEN_CACHE_MAX_CAPACITY</span>
+                    </div>
+                  </td>
+                </tr>
+                <tr>
+                  <td class="text-subtitle-2">Token DB Retry Count</td>
+                  <td>{{ formatValue(settings.registry.token_db_retry_count) }}</td>
+                  <td>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="deep-purple-darken-1" text-color="white"
+                        class="config-chip">TOML</v-chip>
+                      <span class="config-value">registry.token_db_retry_count</span>
+                    </div>
+                    <div class="config-ref">
+                      <v-chip label size="small" color="teal-darken-1" text-color="white"
+                        class="config-chip">ENV</v-chip>
+                      <span class="config-value">KELLNR_REGISTRY__TOKEN_DB_RETRY_COUNT</span>
+                    </div>
+                  </td>
+                </tr>
+                <tr>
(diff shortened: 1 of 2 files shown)


perfectra1n:
Fixed the merge conflict :)

secana:
Hi @perfectra1n! What a great PR, thank you very much!

Would you be so kind and add the new settings to [settings.ts](ui/src/types/settings.ts) and [StartupConfig.vue](ui/src/components/StartupConfig.vue)? This will list the settings in the web UI under "Startup Config" in the settings page. After that I'll merge the PR and release a new version of kellnr.

perfectra1n:
Ahh, forgot those spots. I'll add them there too :)

perfectra1n:
Yeah should be good to go now:
image

PR #956, opened 2026-01-15
ci: Run the CI job only on pull

secana:
Currently the CI job runs on every push to main, which leads to double CI runs as the CI is already ran in the PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,12 +1,8 @@
 name: kellnr-ci
 on:
-  push:
-    branches:
-      - main
   pull_request:
     branches:
       - main
-      - devel
   release:
     types:
       - published


nyurik:
@secana I think you need both:
* one CI run on every change to a pull request - this way developers (including you) get a quick feedback on the state of the PR, and its readiness to merge.  Note that this CI run **skips** release-plz and other non-main steps
* one CI run on every merge to main - this is critical to indicate the state of the main branch, to upload main branch code coverage reports, to run release-plz steps to both generate a release PR and to process that merge PR
* lastly, and this is not as needed - CI run on `release` -- i.e. after a tag is created.  I usually do NOT use this one, but I see you have it. I think it should be removed

P.S. Note that CI run on PR is likely to NOT include the latest merge result between PR and main. In other words, CI checks if the PR branch code is correct, not if the merge result is correct.

secana:
You are right, I re-enabled the CI runs on push and improved the releaze-plz config.

nyurik:
@secana let me know if i can help. I noticed you still have the `run on release` in CI - why do you need it?

PR #958, opened 2026-01-16
ci: refactor YAML syntax for consistency and readability

nyurik:
Also removed unused anchor-yaml-magic stuff.  I have been looking at how to simplify the CI, but this one is just a minor linting PR.

At the moment, there are a lot of manual steps that you may want to modify in the future (TBD, should probably start a new ticket):

* on PR
  * run all validation jobs
* on push to main
  * run the same jobs as run on PR
  * run all packaging jobs
  * run release-plz (could run as a simpler single job after a gate check for all prior steps, see [here](https://github.com/oxibus/automotive_diag/blob/944e7ea6777d97a3c4352e246eab2657ad0166e1/.github/workflows/ci.yml#L74)).  Here, release-plz may want to auto-publish all packages itself, rather than do it in a separate "on-release" trigger.
 
The current per-package `cargo publish` seems wrong, but I might be missing something though.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,8 +26,7 @@ jobs:
         uses: actions/checkout@v6
 
       - uses: cachix/install-nix-action@v31
-        with:
-          nix_path: nixpkgs=channel:nixos-unstable
+        with: { nix_path: 'nixpkgs=channel:nixos-unstable' }
 
       - name: Tests
         run: nix develop --command just test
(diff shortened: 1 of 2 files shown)


secana:
It would be great if `cargo publish --workspace` would work. I tried it and it failed, as the wrong order of packages is published. release-plz uses the same algorithm, so it fails too. I'm not sure where the difficulty for cargo is to get the right order, as I extracted it with a simple script by following the dependencies between the packages.

PR #984, opened 2026-01-26
Performance fix: Reuse reqwest client

Mathis-6:
I noticed that every 5 seconds, CPU time is used for no reason when idling.
Found out it's because a new reqwest::Client is created at every iteration, which is expensive.
This pull request initialize the client once and borrow it.

Before :
before

After :
after


secana:
Thanks! Very welcome improvement.

secana:
Before I can merge, would you be so kind and fix the tests and formatting @Mathis-6?

PR #992, opened 2026-01-26
docs: improve contribution experience

nyurik:
I tried to do the proper "nix" route, and ran into multiple issues. I suspect others may too, so this is more of a WIP PR to optimize contribution experience.

* [x] added link to install nix
* [x] `nix develop` command wouldn't work without extra params. Fixed in docs.
* [x] `nix build` command wouldn't work without extra params. Fixed in docs.
* [ ] `nix build` would only use about 2 cores, making the build extremely slow. Not sure what needs changing.
* [ ] Does nix use regular `cargo build`, or does it try to take over some of that process (similar to Bazel), and possibly result in different behavior?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,14 +45,14 @@ The following tools are needed to build Kellnr:
 
 ### Using Nix
 
-For an easy development environment, you can use the provided `flake.nix`:
+For an easy development environment, you can use the provided `flake.nix`. Install nix from (download page)[https://nixos.org/download/].
 
 ```bash
 # Start a development shell
-nix develop
+nix develop --extra-experimental-features nix-command --extra-experimental-features flakes
 
 # Build the project
-nix build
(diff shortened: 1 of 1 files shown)


nyurik:
i have no idea how nix/flake works, so any recommendations in the docs for devs are up to you.  Overall the fewer the requirements for devs to learn - the better - i.e. if its possible to not have any environment customizations (`cargo build` "just works") - it's ideal.  In my experience that usually works, but a better experience is to have a `just` command to add some optional flags - i.e. `cargo build --all-targets --workspace` is too long to type, so `just build` is better. The next level of annoyance is various apt packages (e.g. `apt install openssl`) - annoying, and we are slowly getting rid of them via `rustls` and the like, but still reality.  And the last level is to require various environments (nix, python venv, docker, bazel, ...) -- they are good to guarantee things, but massive pain for those who don't use them regularly - so if we can avoid them, great.  Just my 2 :coin:

PR #1021, opened 2026-02-02
refactor: simplify user and group mapping using From trait

nyurik:


secana (approved):
Thanks! Much cleaner this way. Can be merged.

PR #1023, opened 2026-02-02
fix: web-ui routing paths when origin.path is used

stevenvdschoot-ly:


secana:
@stevenvdschoot-ly mind having a look at the failing tests?

PR #1024, opened 2026-02-02
chore: run CI on v6

nyurik:


secana (approved):

nyurik:
@secana do you know why things are falling? I might bee at fault for UI, I'll have to investigate, but nix is a bit uncertain

secana:
> @secana do you know why things are falling? I might bee at fault for UI, I'll have to investigate, but nix is a bit uncertain

I pushed a fix for the UI test a few seconds ago. It was just a wrong identifier where "Config" used to get an element, but as there are multiple "Config" strings on the page, it took the wrong one. All tests passed on my machine so I expect them to work in CI, too.

I'll look into the Nix build. I added it to the CI just to make sure the flake.nix is always operational, but never ran it in the CI so far. I would have been more suprised if it worked first try :P

secana:
I guess you can merge this PR. Seeing failures in CI is the exact reason it exists, so there is no need to make all checks pass in here.

PR #1093, opened 2026-02-22
feat(perf): modify storage backend and db queries for performance

perfectra1n:
Hey there, I'm sorry for the huge PR, but I'm hoping that this seriously helps in the long run. When I was really pushing Kellnr to its limits, I found that there were quite a few things that were holding back its performance, and this is a mixture of my notes and changes that followed. 

I know it's a rather large PR and kind of messy, so feel free to let me know if you want me to break it up into smaller pieces, change files, etc...

When many things pull crates simultaneously, especially on slower storage backends like NFS, iSCSI, or S3 over a WAN, Kellnr would choke. Requests would time out, downloads would hang, and the UI would become completely unresponsive (10+ second page loads). The UI freeze was the key clue: it proved this wasn't just "slow storage", but async runtime starvation, download I/O was exhausting all Tokio worker threads, starving every other handler.

## **The…couple of issues found**

Four independent problems, each making the others worse:

1.  The crate cache used manual `get()`\-then-`insert()`. When 50 CI jobs requested the same uncached crate, all 50 independently read it from storage, causing a “thundering herd” lol.
2.  Unlimited concurrent storage reads could saturate I/O, leaving no worker threads for the UI, health checks, or API endpoints. No backpressure
3.  `Cache::new(1000)` counted entries, not bytes. A cache holding 1000 large crates could consume gigabytes; a cache holding 1000 tiny crates would waste capacity.
4.  Despite crate versions being immutable (Kellnr enforces this), download responses had no `Cache-Control` or `ETag` headers. Cargo re-downloaded the same files on every build. No HTTP caching unfortunately.

### **1\. Request coalescing with** `**try_get_with**`

The single highest-impact change. Moka's `try_get_with` guarantees that concurrent requests for the same cache key share a single storage read. The first request triggers the load; all others wait on the same future and receive the result.

Before:

```rust
// Every concurrent request independently reads from storage
if let Some(data) = cache.get(&file_name).await {
    Some(data)
} else {
    let data = self.storage.get(&file_name).await.ok()?.to_vec();
    cache.insert(file_name.clone(), data.clone()).await;
    Some(data)
}

```

After:

```rust
// All concurrent requests for the same key share one storage read
cache
    .try_get_with(file_name, async move {
        storage.get(&key).await
    })
    .await
    .ok()

```

Failed loads are not cached, subsequent requests will retry, which is the correct behavior for transient storage errors.

I also switched from `Vec<u8>` to `bytes::Bytes` throughout the download path. `Bytes` is reference-counted, so cloning it for multiple consumers is a pointer increment rather than a memcpy.

**Files:** `crates/storage/src/cached_crate_storage.rs`, `crates/registry/src/kellnr_api.rs`, `crates/registry/src/cratesio_api.rs`, `crates/registry/Cargo.toml`

### **2\. Byte-weighted cache eviction**

The cache now tracks actual memory usage instead of counting entries. The `cache_size` setting is now interpreted as megabytes (default changed from 1000 entries to 512 MB):

```rust
let max_bytes = settings.registry.cache_size * 1024 * 1024;
Cache::builder()
    .weigher(|_key: &String, value: &Bytes| -> u32 {
        u32::try_from(value.len()).unwrap_or(u32::MAX)
    })
    .max_capacity(max_bytes)
    .build()

```

This is a **semantic change** for existing deployments. Users who had `cache_size = 1000` (meaning 1000 entries) will now get 1000 MB of cache. That's generous but safe, it just means more gets cached. This is of course controlled by the registry cache size in the user's settings.

**Files:** `crates/storage/src/cached_crate_storage.rs`, `crates/settings/src/registry.rs`

### **3\. HTTP caching headers on downloads**

Crate versions are immutable. Once `my-crate-1.0.0` is published, its content never changes. Kellnr now return aggressive caching headers so cargo (and any intermediate proxies/CDNs) can skip re-downloading:

```rust
let etag = format!("\"{package}-{version}\"");
Ok((
    [
        (header::CONTENT_TYPE, HeaderValue::from_static("application/octet-stream")),
        (header::CACHE_CONTROL, HeaderValue::from_static("public, max-age=31536000, immutable")),
        (header::ETAG, HeaderValue::from_str(&etag).unwrap()),
    ],
    file,
).into_response())

```

This applies to both the Kellnr registry download and the crates.io proxy download endpoints.

**Files:** `crates/registry/src/kellnr_api.rs`, `crates/registry/src/cratesio_api.rs`

### **4\. Concurrency limit on download routes**

Download routes now have a configurable concurrency limit via Tower's `ConcurrencyLimitLayer`. Excess requests queue (they wait for a slot) rather than being rejected, so cargo waits instead of failing:

```rust
let mut download_router = Router::new()
    .route("/dl/{package}/{version}/download", get(kellnr_api::download));

if settings.download_max_concurrent > 0 {
    download_router = download_router
        .layer(ConcurrencyLimitLayer::new(settings.download_max_concurrent));
}
if settings.download_timeout_seconds > 0 {
    download_router = download_router
        .layer(TimeoutLayer::with_status_code(
            StatusCode::GATEWAY_TIMEOUT,
            Duration::from_secs(settings.download_timeout_seconds),
        ));
}

```

The timeout layer returns a `504 Gateway Timeout` if a download takes too long (including time spent waiting in the concurrency queue). This prevents a hung storage backend from filling all concurrency slots indefinitely. This is of course configurable via the user's settings.

**Files:** `crates/kellnr/src/routes/kellnr_api_routes.rs`, `crates/kellnr/src/routes/cratesio_api_routes.rs`

### **5\. HTTP client timeouts**

The reqwest `Client` used for crates.io proxy downloads previously had no timeout configured. A slow or unresponsive crates.io would hold connections open indefinitely. I added a shared `build_client()` function with configurable connect and request timeouts, and eliminated duplicate client builder code:

```rust
pub fn build_client(connect_timeout: Duration, request_timeout: Duration) -> Client {
    ClientBuilder::new()
        .gzip(true)
        .connect_timeout(connect_timeout)
        .timeout(request_timeout)
        .default_headers(headers)
        .build()
        .unwrap()
}

```

**Files:** `crates/common/src/cratesio_downloader.rs`, `crates/index/src/cratesio_prefetch_api.rs`

### **6\. S3 storage timeouts**

The `object_store` S3 client was relying on its internal defaults. I now explicitly configure connect and request timeouts from settings:

```rust
let client_options = ClientOptions::new()
    .with_connect_timeout(Duration::from_secs(settings.s3.connect_timeout_seconds))
    .with_timeout(Duration::from_secs(settings.s3.request_timeout_seconds));
s3 = s3.with_client_options(client_options);

```

**Files:** `crates/storage/src/s3_storage.rs`

## **New configuration settings**

All new settings follow existing patterns: configurable via TOML, environment variables (`KELLNR_` prefix, `__` separator), and CLI flags. All have sensible defaults and appear in the settings UI.

| Setting | Default | CLI Flag | Description |
| --- | --- | --- | --- |
| `registry.cache_size` | `512` (was `1000`) | `--registry-cache-size` | Cache size **in MB** (was entry count) |
| `registry.download_timeout_seconds` | `60` | `--registry-download-timeout` | Per-request timeout for downloads (0 = disabled) |
| `registry.download_max_concurrent` | `20` | `--registry-download-max-concurrent` | Max simultaneous download storage reads (0 = unlimited) |
| `proxy.connect_timeout_seconds` | `5` | `--proxy-connect-timeout` | TCP connect timeout for upstream crates.io requests |
| `proxy.request_timeout_seconds` | `30` | `--proxy-request-timeout` | Total request timeout for upstream crates.io downloads |
| `s3.connect_timeout_seconds` | `5` | `--s3-connect-timeout` | TCP connect timeout for S3 operations |
| `s3.request_timeout_seconds` | `30` | `--s3-request-timeout` | Total request timeout for S3 operations |

Environment variable examples:

```
KELLNR_REGISTRY__DOWNLOAD_MAX_CONCURRENT=50
KELLNR_REGISTRY__DOWNLOAD_TIMEOUT_SECONDS=120
KELLNR_S3__REQUEST_TIMEOUT_SECONDS=60

```

## **Why local filesystem (NFS/iSCSI) is safe**

I verified that `object_store`'s `LocalFileSystem` implementation uses `tokio::task::spawn_blocking` for all file I/O. This means slow NFS/iSCSI reads run on the blocking thread pool, not on Tokio worker threads. The concurrency limit and timeout still apply as a backstop, but the async runtime won't be starved by filesystem operations alone.

## **Tests**

*   **10 unit tests** for `CachedCrateStorage` covering: basic get/put, cache hits, request coalescing (50 concurrent requests verify storage is called at most 3 times), cache-disabled mode, error non-caching, cache invalidation on delete, and independent keys.
*   **2 integration tests** for download handlers verifying cache headers (`Cache-Control`, `ETag`, `Content-Type`) and cache hit behavior.
*   All 233 existing workspace tests continue to pass.

## **Files changed (all commits)**

| File | What changed |
| --- | --- |
| `Cargo.toml` | Added `limit` to tower features, `timeout` to tower-http features |
| `crates/kellnr/Cargo.toml` | Added `tower` to runtime dependencies |
| `crates/registry/Cargo.toml` | Added `bytes` dependency |
| `crates/storage/src/cached_crate_storage.rs` | `try_get_with`, `Bytes` type, weigher-based cache, 10 tests |
| `crates/registry/src/kellnr_api.rs` | `Bytes` return type, cache headers, 2 tests |
| `crates/registry/src/cratesio_api.rs` | `Bytes` return type, cache headers |
| `crates/kellnr/src/routes/kellnr_api_routes.rs` | `ConcurrencyLimitLayer` + `TimeoutLayer` on downloads |
| `crates/kellnr/src/routes/cratesio_api_routes.rs` | `ConcurrencyLimitLayer` + `TimeoutLayer` on downloads |
| `crates/settings/src/registry.rs` | `cache_size` default 1000→512 + MB semantics, 2 new fields |
| `crates/settings/src/proxy.rs` | 2 new timeout fields |
| `crates/settings/src/s3.rs` | 2 new timeout fields |
| `crates/settings/src/config_source.rs` | 6 new config source keys |
| `crates/common/src/cratesio_downloader.rs` | `build_client()` with configurable timeouts |
| `crates/index/src/cratesio_prefetch_api.rs` | Uses shared `build_client()`, removed duplicate code |
| `crates/storage/src/s3_storage.rs` | S3 `ClientOptions` with configurable timeouts |
| `crates/storage/tests/s3_tests.rs` | Updated for `Bytes` type |
| `ui/src/types/settings.ts` | New fields in `Registry`, `Proxy`, `S3` types |
| `ui/src/components/StartupConfig.vue` | New config rows, fixed "Cache Size" → "Cache Size (MB)" |
| `scripts/stress-test.sh` | Shell script for load testing download endpoints |

* * *

## Second issue

Under concurrent load, SQLx connection pool acquisition times spiked to 3–4 seconds:

```
WARN sqlx::pool::acquire: acquired connection, but time to acquire exceeded slow threshold
  acquired_after_secs=3.85  slow_acquire_threshold_secs=2.0

```

Every download made 3+ synchronous DB calls just for the download counter (SELECT + 2 UPDATEs) before serving the file. With 50 concurrent downloads and a default pool of ~10 connections, most requests were blocked waiting for a database connection.

The download counter also had a race condition: concurrent requests would read the same `total_downloads` value, each increment it by 1, and write it back, losing all but one increment.

### **Root causes**

1.  Every single download hit the database 3+ times for counter bookkeeping, blocking the response until all writes completed. They were synchronous DB calls.
2.  The counter used `SELECT total_downloads` → compute `value + 1` → `UPDATE total_downloads = new_value`. Under concurrency, multiple requests read the same value and all write `value + 1`, losing increments. Race condition :(
3.  `max_db_connections` defaulted to `0` (library default, typically ~10), far too few for concurrent download workloads. Bumped it up to a few more

#### **1\. In-memory download counter with periodic flush**

Instead of hitting the database on every download, counts accumulate in memory and flush to the database periodically. The download handler now calls `counter.increment()` which takes a `std::sync::Mutex`, increments a `HashMap` entry, and returns, nanoseconds, no DB call, no async, no pool contention.

A background tokio task flushes accumulated counts every `download_counter_flush_seconds` (default 30s). The flush swaps the in-memory map with an empty one (instant), then iterates and writes each unique (crate, version) pair to the database using atomic SQL.

```rust
pub struct DownloadCounter {
    db: Arc<dyn DbProvider>,
    counts: Mutex<HashMap<(NormalizedName, Version), u64>>,
    cached_counts: Mutex<HashMap<(NormalizedName, Version), u64>>,
}

impl DownloadCounter {
    /// Nanosecond increment, no DB call, no async
    pub fn increment(&self, name: NormalizedName, version: Version) {
        let mut counts = self.counts.lock().unwrap();
        *counts.entry((name, version)).or_insert(0) += 1;
    }

    /// Swap map and flush all accumulated counts to DB
    pub async fn flush(&self) {
        let counts = {
            let mut counts = self.counts.lock().unwrap();
            std::mem::take(&mut *counts)
        };
        for ((name, version), count) in counts {
            if let Err(e) = self.db
                .increase_download_counter_by(&name, &version, count).await
            {
                error!("Failed to flush download counter: {e}");
            }
        }
    }
}

```

Trade-off: counts accumulated since the last flush are lost on crash. I figured that this is acceptable, as download counts are analytics, not transactional data.

**Files:** `crates/db/src/download_counter.rs` (new), `crates/db/src/lib.rs`

#### **2\. Atomic SQL for download counters**

All download counter updates now use atomic SQL instead of read-modify-write:

```rust
// Before (race condition):
let total = krate.total_downloads;
krate.total_downloads = Set(total + 1);
krate.update(&self.db_con).await?;

// After (atomic):
krate::Entity::update_many()
    .col_expr(
        krate::Column::TotalDownloads,
        Expr::col(krate::Column::TotalDownloads).add(1),
    )
    .filter(krate::Column::Name.eq(name))
    .exec(&self.db_con)
    .await?;

```

The batched flush path uses `add(count)` for bulk increments. Both the existing per-call methods (`increase_download_counter`) and the new batched methods (`increase_download_counter_by`) use atomic SQL.

**Files:** `crates/db/src/database/mod.rs`, `crates/db/src/provider.rs`

#### **3\. Download handlers decoupled from DB**

Both the Kellnr registry and crates.io proxy download handlers now use the in-memory counter instead of direct DB calls:

```rust
// Before (kellnr_api::download):
if let Err(e) = db.increase_download_counter(&name, &version).await {
    warn!("Failed to increase download counter: {e}");
}

// After:
state.download_counter.increment(name.to_normalized(), version.clone());

```

The crates.io proxy handler previously sent download events through a flume channel to the prefetch thread, which called the DB synchronously per message. It now uses the same in-memory counter.

**Files:** `crates/registry/src/kellnr_api.rs`, `crates/registry/src/cratesio_api.rs`

#### **4\. Wiring and configuration**

The `DownloadCounter` is created in `main.rs`, added to `AppStateData`, and extracted via Axum's `State` in handlers. The flush task is spawned as a background tokio task.

**Files:** `crates/appstate/src/lib.rs`, `crates/kellnr/src/main.rs`, `crates/web-ui/src/user.rs`

### **New configuration settings**

| Setting | Default | CLI Flag | Description |
| --- | --- | --- | --- |
| `registry.download_counter_flush_seconds` | `30` | `--registry-download-counter-flush` | How often to flush in-memory download counts to DB (0 = every download) |
| `registry.max_db_connections` | `50` (was `0`) | `--registry-max-db-connections` | Max database connections in pool (was library default ~10) |

**Files:** `crates/settings/src/registry.rs`, `crates/settings/src/config_source.rs`, `ui/src/types/settings.ts`, `ui/src/components/StartupConfig.vue`

### **Supporting changes**

*   `**Hash**` **for** `**NormalizedName**`: Added `Hash` derive, needed for `HashMap` keys in the counter.
*   `**Hash**` **for** `**Version**`: Custom implementation consistent with the existing custom `PartialEq` (which normalizes via semver parsing). Falls back to string hashing for invalid semver strings.

**Files:** `crates/common/src/normalized_name.rs`, `crates/common/src/version.rs`

### **Files changed**

| File | What changed |
| --- | --- |
| `crates/db/src/download_counter.rs` | New: in-memory counter with periodic flush |
| `crates/db/src/lib.rs` | Added `download_counter` module |
| `crates/db/src/provider.rs` | Added `increase_*_counter_by` trait methods + mock stubs |
| `crates/db/src/database/mod.rs` | Atomic SQL for all counter methods, implemented `_by` variants |
| `crates/appstate/src/lib.rs` | Added `download_counter` to `AppStateData`, `DownloadCounterState` type alias |
| `crates/kellnr/src/main.rs` | Counter initialization, flush task spawn |
| `crates/registry/src/kellnr_api.rs` | Replaced DB call with `counter.increment()` |
| `crates/registry/src/cratesio_api.rs` | Replaced channel send with `counter.increment_cached()` |
| `crates/settings/src/registry.rs` | New `download_counter_flush_seconds`, changed `max_db_connections` default |
| `crates/settings/src/config_source.rs` | New config source key |
| `crates/common/src/normalized_name.rs` | Added `Hash` derive |
| `crates/common/src/version.rs` | Custom `Hash` implementation |
| `crates/web-ui/src/user.rs` | Updated test helper for new `AppStateData` field |
| `ui/src/types/settings.ts` | New field in `Registry` type |
| `ui/src/components/StartupConfig.vue` | New config row for flush interval |

### DIFF
--- a/crates/db/src/download_counter.rs
+++ b/crates/db/src/download_counter.rs
@@ -14,8 +14,8 @@ use crate::DbProvider;
 /// 2-calls-per-unique-crate-per-flush.
 pub struct DownloadCounter {
     db: Arc<dyn DbProvider>,
-    counts: Mutex<HashMap<(NormalizedName, Version), u64>>,
-    cached_counts: Mutex<HashMap<(NormalizedName, Version), u64>>,
+    counts: Mutex<HashMap<(NormalizedName, String), u64>>,
+    cached_counts: Mutex<HashMap<(NormalizedName, String), u64>>,
     flush_interval: u64,
 }
 
@@ -37,7 +37,7 @@ impl DownloadCounter {
                 warn!("Failed to increment download counter for {name} {version}: {e}");
             }
         } else {
-            self.increment(name, version);
+            self.increment(name, &version);
         }
     }
 
@@ -53,23 +53,23 @@ impl DownloadCounter {
                 warn!("Failed to increment cached download counter for {name} {version}: {e}");
             }
         } else {
-            self.increment_cached(name, version);
+            self.increment_cached(name, &version);
         }
     }
 
     /// Record a download for a kellnr-hosted crate. Instant, no DB call.
-    pub fn increment(&self, name: NormalizedName, version: Version) {
+    fn increment(&self, name: NormalizedName, version: &Version) {
         let mut counts = self.counts.lock().expect("download counter lock poisoned");
-        *counts.entry((name, version)).or_insert(0) += 1;
+        *counts.entry((name, version.to_string())).or_insert(0) += 1;
     }
 
     /// Record a download for a cached crates.io crate. Instant, no DB call.
-    pub fn increment_cached(&self, name: NormalizedName, version: Version) {
+    fn increment_cached(&self, name: NormalizedName, version: &Version) {
         let mut counts = self
             .cached_counts
             .lock()
             .expect("cached download counter lock poisoned");
-        *counts.entry((name, version)).or_insert(0) += 1;
+        *counts.entry((name, version.to_string())).or_insert(0) += 1;
     }
 
     /// Flush all accumulated counts to the database.
@@ -96,6 +96,7 @@ impl DownloadCounter {
 
         // Flush kellnr crate counts
         for ((name, version), count) in counts {
+            let version = Version::from_unchecked_str(&version);
             if let Err(e) = self
                 .db
                 .increase_download_counter_by(&name, &version, count)
@@ -107,6 +108,7 @@ impl DownloadCounter {
 
         // Flush cached crates.io crate counts
         for ((name, version), count) in cached_counts {
+            let version = Version::from_unchecked_str(&version);
             if let Err(e) = self
                 .db
                 .increase_cached_download_counter_by(&name, &version, count)


perfectra1n:
I've been staring at this code for a very, very long time so it's possible that I missed something silly e.g. a setting entry missing on the UI, so feel free to let me know anything silly I did 😂

crates/kellnr/src/main.rs
```diff
@@ -177,6 +178,23 @@ async fn run_server(settings: Settings) {
     // Initialize OAuth2/OIDC handler if enabled
     let oauth2_handler = init_oauth2_handler(&settings).await;
 
+    // Initialize download counter with periodic flush
```
secana, line 182: I think there is a bug in the `download_counter_flush_seconds = 0` logic.
The setting doc says "0 = flush every download", but the implementation does no such thing. In crates/kellnr/src/main.rs:189-196, when flush_interval is 0, no background task is spawned. And increment()
never calls flush() inline. So setting this to 0 means counters accumulate in memory forever and are never written to the database.
If that is correct, implement synchronous-per-download flushing when flush_interval == 0 (call db.increase_download_counter() directly in the handler, like the old behavior).

crates/common/src/version.rs
```diff
@@ -99,6 +100,17 @@ impl PartialEq for Version {
     }
 }
 
+impl Hash for Version {
```
secana: The is `PartialEq` inconsistency for `Version`. crates/common/src/version.rs — The existing PartialEq (line 94-99) calls `.unwrap()` on `semver::Version::parse()`, so it panics on invalid semver. The new Hash impl (added at line 142-151) gracefully falls back to string hashing for invalid semver.

The Hash/Eq contract (a == b implies hash(a) == hash(b)) is technically upheld because PartialEq panics before it could return true for invalid semver. But this is fragile — the two implementations have different assumptions about what inputs are valid. If from_`unchecked_str` is used with invalid semver, you get a type that can be hashed but panics on comparison. IMHO it does not make sense to return a hash for an invalid semver.

crates/registry/src/kellnr_api.rs
```diff
@@ -592,20 +592,35 @@ pub async fn download(
...
+                [
+                    (
+                        header::CONTENT_TYPE,
+                        HeaderValue::from_static("application/octet-stream"),
+                    ),
+                    (
+                        header::CACHE_CONTROL,
+                        HeaderValue::from_static("public, max-age=31536000, immutable"),
+                    ),
+                    (header::ETAG, HeaderValue::from_str(&etag).unwrap()),
```
secana: Don't `unwarp()` in a route handler. Use proper error handling and return an appropriate error HTTP error if an ETAG is in the wrong format.

crates/registry/src/cratesio_api.rs
```diff
@@ -114,8 +108,25 @@ pub async fn download(
...
+        [
+            (
+                header::CONTENT_TYPE,
+                HeaderValue::from_static("application/octet-stream"),
+            ),
+            (
+                header::CACHE_CONTROL,
+                HeaderValue::from_static("public, max-age=31536000, immutable"),
+            ),
+            (header::ETAG, HeaderValue::from_str(&etag).unwrap()),
```
secana: Don't `unwarp()` in a route handler. Use proper error handling and return an appropriate error HTTP error if an ETAG is in the wrong format.

crates/storage/src/s3_storage.rs
```diff
@@ -81,6 +84,13 @@ impl TryFrom<(&str, &Settings)> for S3Storage {
         if let Some(secret_key) = &settings.s3.secret_key {
             s3 = s3.with_secret_access_key(secret_key);
         }
+
+        let client_options = ClientOptions::new()
+            .with_connect_timeout(Duration::from_secs(settings.s3.connect_timeout_seconds))
+            .with_timeout(Duration::from_secs(settings.s3.request_timeout_seconds))
+            .with_allow_http(settings.s3.allow_http);
```
secana: Line 70 already calls `.with_allow_http(settings.s3.allow_http)` on the AmazonS3Builder. It may work, but the duplication is confusing and the interaction should be verified. If `ClientOptions.with_allow_http` overrides the builder setting, the builder call becomes dead code. If they're both needed, a comment explaining why would help.

crates/db/src/download_counter.rs
```diff
@@ -0,0 +1,95 @@
...
+                .increase_cached_download_counter_by(&name, &version, count)
+                .await
+            {
+                warn!(
+                    "Failed to flush cached download counter for {name} {version} (count={count}): {e}"
+                );
+            }
+        }
+
+        info!(
```
secana: Don't log `info!` in each loop iteration. Use `trace!` if the log is needed.

crates/registry/src/kellnr_api.rs
```diff
@@ -2394,6 +2409,134 @@ mod reg_api_tests {
...
+        let content_type = r.headers().get(header::CONTENT_TYPE).unwrap();
+        assert_eq!(content_type, "application/octet-stream");
+
+        // Verify the body is non-empty
+        let body = r.into_body().collect().await.unwrap().to_bytes();
+        assert!(!body.is_empty());
+    }
+
+    #[tokio::test]
+    async fn download_second_call_uses_cache() {
```
secana: The test downloads twice and verifies the content matches, then does a third download to check it returns 200. But it doesn't verify that the second/third calls came from cache rather than storage (no call-count tracking). Compare with the CachedCrateStorage unit tests which properly track storage call counts via CountingStorage.

secana:
Hi @perfectra1n! Thanks for the PR. Performance improvements are very welcome. As this is a huge PR I need some time to review it. Here are some first impressions:

When Kellnr shuts down, up to 30 seconds of accumulated download counts are silently lost. The PR acknowledges this tradeoff ("download counts are analytics, not transactional data"), which is easonable, but a shutdown hook calling counter.flush().await before exiting would be a low-effort improvement. Tokio's signal::ctrl_c() or a CancellationToken pattern could handle this.

The PR description says the batch flush reduces DB calls from "3-calls-per-download to 2-calls-per-unique-crate-per-flush". But increase_download_counter_by still does 3 queries: 

1. update_many for total_downloads
2. find().one() to get crate_id
3. update_many for version downloads

This could be reduced to 2 queries if crate_id were looked up via a subquery or JOIN in the version update.

In DownloadCounter::flush(), if the total_downloads update succeeds but the version-specific update fails, the counts become inconsistent (total_downloads is incremented but the version's count is not). Wrapping the two updates in a transaction for each (name, version) pair would be more correct.

Both kellnr_api_routes.rs and cratesio_api_routes.rs use settings.registry.download_max_concurrent. This means the 20-slot default is shared per-route-type but not globally. If 20 kellnr downloads and 20 crates.io downloads happen simultaneously, you get 40 concurrent storage reads, not 20. That comes unexpected from a users perspective.

Crates are immutable on crates.io but not on Kellnr. You can delete a version and re-upload it with different content. So you would need to take the content of the crate into the hash. Kellnr sets the ETAG for all cargo pre-fetches already. Why not reuse the already present etag, if that helps with CDN caching the request.

BREAKING: Going from number of cached crates to a fixed cache size is a breaking changes that I would not like to have. Please revert that change.

BREAKING: `registry.max_db_connections = 50` is a breaking change as it changes the behavior of the default. Please revert. The description assumes that `0` is a low limit but `0` means the limit is disabled. So setting it to `50` does not increase but decrease it.

The PR description mentions a scripts/stress-test.sh file but it doesn't appear in the diff. Either it wasn't committed or it was removed before pushing.

secana:
The PR feels very vibe coded. I'm not against using AI to code, quite the opposite but it came out as a huge mixed bag of several fixes which makes it hard to review and reason about. @nyurik Can you help with reviewing or shall @perfectra1n split the PR?

crates/storage/src/cached_crate_storage.rs
```diff
@@ -107,3 +115,330 @@ impl CachedCrateStorage {
...
+            Self { data, metrics }
+        }
+    }
+
+    #[async_trait]
+    impl Storage for CountingStorage {
+        async fn get(&self, key: &str) -> Result<Bytes, StorageError> {
+            self.metrics.get_count.fetch_add(1, Ordering::SeqCst);
+            // Simulate slow storage
+            tokio::time::sleep(std::time::Duration::from_millis(50)).await;
```
secana: Do not `sleep` in tests. Use a deterministic sync gate if needed.

crates/registry/src/cratesio_api.rs
```diff
@@ -85,21 +84,16 @@ pub async fn search(params: SearchParams) -> ApiResult<String> {
...
 
-    if let Some(file) = crate_storage.get(&name, &version).await {
-        let msg = DownloadData {
-            name: name.into(),
-            version,
-        };
-        if let Err(e) = sender.send(CratesioPrefetchMsg::IncDownloadCnt(msg)) {
-            warn!("Failed to send IncDownloadCnt message: {e}");
-        }
+    let file = if let Some(file) = crate_storage.get(&name, &version).await {
```
secana, line 92: The in-memory counter increments only on cache hits; the first uncached download fetches from upstream and returns without incrementing. This loses the initial download count for every crate/version.

crates/index/src/cratesio_prefetch_api.rs
```diff
@@ -26,16 +26,10 @@ use super::config_json::ConfigJson;
...
-    headers.insert(
-        reqwest::header::USER_AGENT,
-        reqwest::header::HeaderValue::from_static("kellnr.io/kellnr"),
-    );
-    ClientBuilder::new()
-        .gzip(true)
-        .default_headers(headers)
-        .build()
-        .unwrap()
+    kellnr_common::cratesio_downloader::build_client(
```
secana: The client is still built with fixed 5s/30s timeouts instead of proxy.connect_timeout_seconds and proxy.request_timeout_seconds. The settings are introduced in crates/settings/src/proxy.rs but not applied for prefetch.

crates/kellnr/src/routes/cratesio_api_routes.rs
```diff
@@ -1,19 +1,47 @@
...
+        "/dl/{package}/{version}/download",
+        get(cratesio_api::download),
+    );
+
+    if settings.download_max_concurrent > 0 {
+        download_router =
+            download_router.layer(ConcurrencyLimitLayer::new(settings.download_max_concurrent));
+    }
+    if settings.download_timeout_seconds > 0 {
+        download_router = download_router.layer(TimeoutLayer::with_status_code(
```
secana: For large crates or slow clients can get 504 even when storage is fine, due to the timeout limit. As the timeout is a setting, maybe print a warning that tells the user what happend such that he can increase the limit.

crates/kellnr/src/routes/kellnr_api_routes.rs
```diff
@@ -18,6 +25,25 @@ pub fn create_routes(state: AppStateData, max_crate_size: usize) -> OpenApiRoute
...
+        "/dl/{package}/{version}/download",
+        get(kellnr_api::download),
+    );
+
+    if settings.download_max_concurrent > 0 {
+        download_router =
+            download_router.layer(ConcurrencyLimitLayer::new(settings.download_max_concurrent));
+    }
+    if settings.download_timeout_seconds > 0 {
+        download_router = download_router.layer(TimeoutLayer::with_status_code(
```
secana: For large crates or slow clients can get 504 even when storage is fine, due to the timeout limit. As the timeout is a setting, maybe print a warning that tells the user what happend such that he can increase the limit.

perfectra1n:
Fair enough - I'll resolve the comments you've made, and then I'll split up this PR/branch into multiple smaller ones so it's easier to digest :)

perfectra1n:
Thanks for going through it still, even though it's such a mess! I'm actually quite surprised how much more "pressure" Kellnr can handle with the changes in this PR, so I promise it's worth it :)

PR #1094, opened 2026-02-23
Fix `max_crate_size` default value

wojciech-graj:
max_crate_size is in units of MB, but the comment and default value incorrectly assumed KB.

Usage: https://github.com/kellnr/kellnr/blob/dd24826e91ab09bcac578f0f00946ab819e22a13/crates/kellnr/src/routes/kellnr_api_routes.rs#L17
The docs (https://kellnr.io/documentation) also state that it is in MB.


secana:
You are right. Good catch!

PR #1097, opened 2026-02-25
feat(storage): cache request coalescing with `try_get_with` and `Bytes` type

perfectra1n:
This is a continuation of #1093, split up into 4 PRs. The requested changes from #1093 have also been implemented. This is the 1st PR.

This PR addresses a performance issue where multiple concurrent requests for the same crate were causing redundant storage reads. The solution implements request coalescing using moka's `try_get_with` method and switches from `Vec<u8>` to `Bytes` for more efficient memory handling of crate data.

The core change is in the cached storage layer. Previously, when multiple requests arrived simultaneously for a crate that wasn't cached, each request would independently fetch from storage. Now the implementation uses `try_get_with`, which ensures that only one fetch operation occurs even when many requests arrive concurrently. When one request initiates a fetch, subsequent requests for the same key wait for the first one to complete and share its result.

Here's the key change in `cached_crate_storage.rs`:

```rust
pub async fn get(&self, name: &OriginalName, version: &Version) -> Option<Bytes> {
    let file_name = Self::file_name(name, version);
    match self.cache {
        Some(ref cache) => {
            let storage = &self.storage;
            let key = file_name.clone();
            cache
                .try_get_with(file_name, async move { storage.get(&key).await })
                .await
                .ok()
        }
        None => self.storage.get(&file_name).await.ok(),
    }
}

```

The old approach would check the cache with `cache.get()`, and if it missed, would fetch from storage and then manually insert into the cache. This created a race condition where concurrent requests could all miss the cache and each initiate their own storage fetch. The new approach delegates this logic to moka's coalescing mechanism.

The switch from `Vec<u8>` to `Bytes` provides several benefits. `Bytes` is a reference-counted buffer type that allows multiple consumers to share the same underlying data without copying. When serving downloads to many concurrent clients, this means the crate data can be shared efficiently in memory rather than being cloned for each response. The change propagates through the download endpoints where the old code returned `Ok(file)` directly but now returns `Ok(file.to_vec())` to convert the `Bytes` back to `Vec<u8>` for the response body.

The PR includes comprehensive tests that verify the coalescing behavior. One test launches 50 concurrent requests for the same crate and verifies that storage is called at most 3 times rather than 50 times. The test uses a counting storage backend that tracks how many times `get()` is called, and it gates the storage responses with an atomic boolean to ensure all requests queue up before the fetch completes. This demonstrates that the coalescing actually works under concurrent load.

Another important test verifies that failed storage reads are not cached. If a crate doesn't exist and multiple requests try to fetch it, each request should get `None` and the absence should not be cached. The test calls `get()` twice for a missing crate and confirms that storage is called both times, proving that errors don't get memoized.

The cache builder now uses explicit builder syntax instead of the convenience constructor to maintain compatibility with moka's API:

```rust
let cache = if settings.registry.cache_size > 0 {
    Some(
        Cache::builder()
            .max_capacity(settings.registry.cache_size)
            .build(),
    )
} else {
    None
};

```

This change has immediate impact on Kellnr's performance under load. When a popular crate is requested by many clients simultaneously, the server will only perform one disk or S3 read instead of N reads. The `Bytes` type reduces memory pressure by allowing the same buffer to be referenced by multiple responses. The combination makes Kellnr significantly more efficient when serving high-demand crates.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,7 @@
           version = "0.1.0";
           src = ./ui;
 
-          npmDepsHash = "sha256-peTxkte1wVK48rjy+EoaepKNgXH7M3m3N6PHMYAMBQ4=";
+          npmDepsHash = "sha256-fVa+RP5SGqopGvh5XlhbUtXcKAW3w8pegyoeiOfDZFI=";
 
           buildPhase = ''
             npm run build


crates/registry/src/cratesio_api.rs
```diff
@@ -99,7 +99,7 @@ pub async fn download(
             warn!("Failed to send IncDownloadCnt message: {e}");
         }
 
-        Ok(file)
+        Ok(file.to_vec())
```
secana: `to_vec()` copies the data which takes away a lot of the gained performance through using `Bytes` as a data type. 

cratesio_api.rs and kellnr_api.rs should use `Bytes` in the API to instead of a vector. That would make the code cleaner and faster as no conversion is needed.

@perfectra1n Can you try to return `Bytes` from the API? This would make your PR faster as allocation is reduced.

crates/registry/src/kellnr_api.rs
```diff
@@ -605,7 +605,7 @@ pub async fn download(
     }
 
     match cs.get(&package, &version).await {
-        Some(file) => Ok(file),
+        Some(file) => Ok(file.to_vec()),
```
secana: See comment above

secana (changes requested):
Have a look at the comment. Besides that, the PR looks good to me.

PR #1098, opened 2026-02-25
feat(db): in-memory download counter with periodic flush and atomic SQL

perfectra1n:
This is a continuation of #1093, split up into 4 PRs. The requested changes from #1093 have also been implemented. This is the 2nd PR.

This PR redesigns the download counter system to eliminate database race conditions and reduce database load by batching updates. The solution introduces an in-memory accumulator that periodically flushes to the database, and fixes a critical race condition in the SQL update logic.

The old implementation had a read-modify-write race condition. When a download was recorded, the code would read the current download count from the database, increment it in memory, and write it back. If two downloads happened concurrently, they could both read the same value, increment it, and write back the same result, losing one of the counts. Here's what the old code looked like:

```rust
let krate = self.get_krate_model(crate_name).await?;
let crate_id = krate.id;
let crate_total_downloads = krate.total_downloads;

let mut k: krate::ActiveModel = krate.into();
k.total_downloads = Set(crate_total_downloads + 1);
k.update(&self.db_con).await?;

```

The new implementation uses atomic SQL updates that perform the increment directly in the database:

```rust
krate::Entity::update_many()
    .col_expr(
        krate::Column::TotalDownloads,
        Expr::col(krate::Column::TotalDownloads).add(1),
    )
    .filter(krate::Column::Name.eq(crate_name))
    .exec(&self.db_con)
    .await?;

```

This generates SQL like `UPDATE krate SET total_downloads = total_downloads + 1 WHERE name = ?`, which is atomic at the database level. No matter how many concurrent updates occur, the database will serialize them correctly and each increment is guaranteed to apply.

The new system also introduces batching to reduce database pressure. Every download increments an in-memory counter rather than hitting the database immediately. A background task wakes up every 30 seconds (configurable) and flushes all accumulated counts to the database in batch. Here's the new `DownloadCounter` structure:

```rust
pub struct DownloadCounter {
    db: Arc<dyn DbProvider>,
    counts: Mutex<HashMap<(NormalizedName, String), u64>>,
    cached_counts: Mutex<HashMap<(NormalizedName, String), u64>>,
    flush_interval: u64,
}

```

The flush operation swaps out the accumulated maps with empty ones to minimize lock contention, then writes all the counts in a loop. For each unique crate version that was downloaded, it calls a new database method `increase_download_counter_by` that takes a count parameter:

```rust
async fn increase_download_counter_by(
    &self,
    crate_name: &NormalizedName,
    crate_version: &Version,
    count: u64,
) -> DbResult<()> {
    let txn = self.db_con.begin().await?;

    krate::Entity::update_many()
        .col_expr(
            krate::Column::TotalDownloads,
            Expr::col(krate::Column::TotalDownloads).add(count as i64),
        )
        .filter(krate::Column::Name.eq(crate_name))
        .exec(&txn)
        .await?;

    // ... update version-specific counter as well ...

    txn.commit().await?;
    Ok(())
}

```

This method uses a transaction to ensure both the total crate downloads and the version-specific downloads are updated atomically together. If 15 downloads happened for version 1.0.0 during the flush interval, this will add 15 to both counters in a single database round-trip.

The system supports a flush interval of 0 for deployments that want immediate consistency. When the interval is 0, every download goes directly to the database using the atomic SQL update. This provides a migration path and allows operators to choose between throughput and real-time accuracy.

The main function sets up the periodic flush task:

```rust
let flush_interval = settings.registry.download_counter_flush_seconds;
let download_counter = Arc::new(DownloadCounter::new(db.clone(), flush_interval));
if flush_interval > 0 {
    let counter = download_counter.clone();
    info!("Starting download counter flush task (interval: {flush_interval}s)");
    tokio::spawn(async move {
        let mut interval = tokio::time::interval(Duration::from_secs(flush_interval));
        interval.tick().await; // skip first immediate tick
        loop {
            interval.tick().await;
            counter.flush().await;
        }
    });
}

```

The code also ensures that accumulated counts are flushed on shutdown. The main server loop wraps the axum serve in a `tokio::select!` that listens for Ctrl-C, and when shutdown is triggered, it calls `flush()` one final time before exiting. This prevents losing download counts that were accumulated in memory but not yet written to the database.

The download endpoints are simplified because they no longer need to handle database errors inline. The old code would log a warning if the database update failed, but otherwise continue serving the file. The new code just calls `increment_and_maybe_flush()` which handles the flush decision internally:

```rust
download_counter
    .increment_and_maybe_flush(name.to_normalized(), version.clone())
    .await;

```

This PR transforms download counting from a source of database contention into a low-overhead operation. For a busy server handling 1000 downloads per minute across 50 unique crate versions, the old system would perform 3000 database calls per minute. The new system performs 100 database calls per minute (50 unique versions × 2 tables, every 30 seconds). The atomic SQL fixes ensure that even with high concurrency, counts remain accurate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4015,6 +4015,7 @@ name = "kellnr-registry"
 version = "6.0.3"
 dependencies = [
  "axum",
+ "bytes",
  "chrono",
  "cookie",
  "http-body-util",
--- a/crates/registry/Cargo.toml
+++ b/crates/registry/Cargo.toml
@@ -32,6 +32,7 @@ serde_json.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
 url.workspace = true
+bytes.workspace = true
 
 [dev-dependencies]
 cookie.workspace = true
--- a/crates/registry/src/cratesio_api.rs
+++ b/crates/registry/src/cratesio_api.rs
@@ -2,7 +2,8 @@ use axum::extract::{Path, Request, State};
 use axum::http::StatusCode;
 use axum::middleware::Next;
 use axum::response::Response;
-use kellnr_appstate::{CrateIoStorageState, DownloadCounterState, SettingsState};
+use bytes::Bytes;
+use kellnr_appstate::{CrateIoStorageState, CratesIoPrefetchSenderState, DownloadCounterState, SettingsState};
 use kellnr_common::cratesio_downloader::{CLIENT, download_crate};
 use kellnr_common::original_name::OriginalName;
 use kellnr_common::version::Version;
@@ -86,7 +87,7 @@ pub async fn download(
     State(crate_storage): CrateIoStorageState,
     State(download_counter): DownloadCounterState,
     State(settings): SettingsState,
-) -> Result<Vec<u8>, StatusCode> {
+) -> Result<Bytes, StatusCode> {
     trace!("Downloading crate: {name} ({version})");
 
     let file = if let Some(file) = crate_storage.get(&name, &version).await {
--- a/crates/registry/src/kellnr_api.rs
+++ b/crates/registry/src/kellnr_api.rs
@@ -14,6 +14,7 @@ use kellnr_common::search_result::{Crate, SearchResult};
 use kellnr_common::version::Version;
 use kellnr_common::webhook::WebhookEvent;
 use kellnr_db::DbProvider;
+use bytes::Bytes;
 use kellnr_error::api_error::{ApiError, ApiResult};
 
 use crate::pub_data::{EmptyCrateData, PubData};
@@ -591,7 +592,7 @@ pub async fn download(
     State(state): AppState,
     token: token::OptionToken,
     Path((package, version)): Path<(OriginalName, Version)>,
-) -> ApiResult<Vec<u8>> {
+) -> ApiResult<Bytes> {
     let db = state.db;
     let cs = state.crate_storage;
     let download_counter = state.download_counter;
--- a/crates/storage/src/cached_crate_storage.rs
+++ b/crates/storage/src/cached_crate_storage.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 use std::sync::Arc;
 
+use bytes::Bytes;
 use kellnr_common::original_name::OriginalName;
 use kellnr_common::version::Version;
 use kellnr_settings::Settings;
@@ -9,7 +10,7 @@ use moka::future::Cache;
 use crate::storage::Storage;
 use crate::storage_error::StorageError;
 
-pub type CrateCache = Cache<String, Vec<u8>>;
+pub type CrateCache = Cache<String, Bytes>;
 pub type DynStorage = Box<dyn Storage + Send + Sync>;
 
 pub struct CachedCrateStorage {
@@ -21,7 +22,11 @@ pub struct CachedCrateStorage {
 impl CachedCrateStorage {
     pub fn new(settings: &Settings, storage: DynStorage) -> Self {
         let cache = if settings.registry.cache_size > 0 {
-            Some(Cache::new(settings.registry.cache_size))
+            Some(
+                Cache::builder()
+                    .max_capacity(settings.registry.cache_size)
+                    .build(),
+            )
         } else {
             None
         };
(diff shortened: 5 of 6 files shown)


crates/kellnr/src/main.rs
```diff
@@ -177,6 +178,25 @@ async fn run_server(settings: Settings) {
     // Initialize OAuth2/OIDC handler if enabled
     let oauth2_handler = init_oauth2_handler(&settings).await;
 
+    // Initialize download counter with periodic flush
+    let flush_interval = settings.registry.download_counter_flush_seconds;
+    let download_counter = Arc::new(DownloadCounter::new(db.clone(), flush_interval));
+    if flush_interval > 0 {
+        let counter = download_counter.clone();
+        info!("Starting download counter flush task (interval: {flush_interval}s)");
```
secana: please use `trace!` instead of `info!`

crates/kellnr/src/main.rs
```diff
@@ -209,12 +230,26 @@ async fn run_server(settings: Settings) {
...
-    axum::serve(listener, app).await.unwrap();
+
+    tokio::select! {
+        result = axum::serve(listener, app) => {
+            if let Err(e) = result {
+                error!("Server error: {e}");
+            }
+        }
+        _ = tokio::signal::ctrl_c() => {
+            info!("Shutdown signal received, flushing download counters...");
```
secana: please use `trace!` instead of `info!`

crates/kellnr/src/main.rs
```diff
@@ -209,12 +230,26 @@ async fn run_server(settings: Settings) {
...
+        }
+    }
+
+    // Flush any accumulated download counts before exiting
+    download_counter_for_shutdown.flush().await;
+    info!("Download counters flushed. Shutting down.");
```
secana: please use trace! instead of info!

crates/kellnr/src/main.rs
```diff
@@ -209,12 +230,26 @@ async fn run_server(settings: Settings) {
...
     info!("Kellnr has been started on http://{addr}/");
-    axum::serve(listener, app).await.unwrap();
+
+    tokio::select! {
+        result = axum::serve(listener, app) => {
+            if let Err(e) = result {
+                error!("Server error: {e}");
+            }
+        }
+        _ = tokio::signal::ctrl_c() => {
```
secana, line 255: Can be improved further to support signals by Docker for the Docker image:

```rust
  #[cfg(unix)]
  let sigterm = async {
      tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
          .expect("failed to install SIGTERM handler")
          .recv()
          .await;
  };
  #[cfg(not(unix))]
  let sigterm = std::future::pending::<()>();

  tokio::select! {
      result = axum::serve(listener, app) => { ... }
      _ = tokio::signal::ctrl_c() => { ... }
      _ = sigterm => { ... }
  }

File: crates/db/src/download_counter.rs [hunk 5]
```diff
@@ -0,0 +1,127 @@
...
+
+        // Flush kellnr crate counts
+        for ((name, version), count) in counts {
+            let version = Version::from_unchecked_str(&version);
+            if let Err(e) = self
+                .db
+                .increase_download_counter_by(&name, &version, count)
+                .await
+            {
+                warn!("Failed to flush download counter for {name} {version} (count={count}): {e}");
```
secana, line 104: Just an idea: Currently the count is silently dropped when the insert fails for some reason. That should not happen in the happy path and even if the count is lost, it's not a huge issue. But instead of dropping the count it could be re-inserted on failure such that the next insert in the DB can retry to add the count.

crates/db/src/database/mod.rs
```diff
@@ -472,16 +472,25 @@ impl DbProvider for Database {
         crate_name: &NormalizedName,
         crate_version: &Version,
     ) -> DbResult<()> {
-        let krate = self.get_krate_model(crate_name).await?;
-        let crate_id = krate.id;
-        let crate_total_downloads = krate.total_downloads;
+        // Atomic update for crate total_downloads (fixes race condition)
```
secana: The new increase_download_counter_by correctly wraps both the crate-level and version-level updates in a transaction. But the existing increase_download_counter (used when flush_interval=0) was only
updated to use atomic SQL — it still does two separate non-transactional queries. This inconsistency means the flush_interval=0 path can still partially fail (crate total updated, version count not).
Should use a transaction there too.

secana:
The old download counting for cached crates went through the prefetch channel (CratesioPrefetchMsg::IncDownloadCnt). The PR removes the sender side in cratesio_api.rs but the receiver in cratesio_prefetch_api.rs:445-454 still handles IncDownloadCnt messages and calls db.increase_cached_download_counter(). The DownloadData struct and IncDownloadCnt variant in cratesio_prefetch_msg.rs are also now dead code. The dead code needs to be cleaned up.

crates/db/src/download_counter.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::DbProvider;
+
+/// In-memory download counter with periodic DB flush.
+///
+/// Accumulates download counts in memory and writes them in batch,
+/// reducing DB pressure from 3-calls-per-download to
+/// 2-calls-per-unique-crate-per-flush.
+pub struct DownloadCounter {
+    db: Arc<dyn DbProvider>,
+    counts: Mutex<HashMap<(NormalizedName, String), u64>>,
```
secana: The counter maps use (NormalizedName, String) as keys, with version.to_string() on insert and Version::from_unchecked_str() on flush. Since Hash was already added to NormalizedName in this PR, adding Hash to Version as well would let the map use (NormalizedName, Version) directly and avoid the unchecked round-trip.

secana:
PR #1097 (the companion PR) added thorough tests for its changes. This PR doesn't include any unit tests for DownloadCounter — the flush logic, the flush_interval=0 fallback, and the concurrent increment behavior are all untested. A few tests with a mock DbProvider would be valuable.

secana (changes requested):

PR #1099, opened 2026-02-25
feat(http): add Cache-Control and ETag headers to download endpoints

perfectra1n:
This is a continuation of #1093, split up into 4 PRs. The requested changes from #1093 have also been implemented. This is the 3rd PR.

This PR adds proper HTTP caching headers to crate download endpoints, allowing clients and CDNs to cache crate files effectively. Crate files are immutable once published, so they can be cached aggressively without worrying about stale data. The implementation adds Cache-Control and ETag headers to both kellnr-hosted crates and proxied crates.io downloads.

For kellnr-hosted crates, the implementation computes a SHA-256 hash of the crate file content and uses it as the ETag. This provides strong validation because any change to the file would produce a different hash. Here's the change to the kellnr download endpoint:

```rust
pub async fn download(
    State(state): AppState,
    token: token::OptionToken,
    Path((package, version)): Path<(OriginalName, Version)>,
) -> Result<Response, ApiError> {
    // ... auth and counter logic ...
    
    match cs.get(&package, &version).await {
        Some(file) => {
            let hash = sha256::digest(&*file);
            let etag = format!("\"{hash}\"");
            Ok((
                [
                    (header::CONTENT_TYPE, HeaderValue::from_static("application/octet-stream")),
                    (header::CACHE_CONTROL, HeaderValue::from_static("public, must-revalidate")),
                    (header::ETAG, HeaderValue::from_str(&etag).map_err(|_| {
                        ApiError::new("Invalid ETag value", "", StatusCode::INTERNAL_SERVER_ERROR)
                    })?),
                ],
                file,
            ).into_response())
        }
        None => Err(RegistryError::CrateNotFound.into()),
    }
}

```

The Cache-Control header uses `public, must-revalidate`. The `public` directive allows shared caches like CDNs to store the response. The `must-revalidate` directive tells caches that they must revalidate with the origin server once the cached response expires, rather than serving stale content. This strikes a balance between cacheability and ensuring clients don't get outdated versions if a crate is somehow republished with the same version number.

For crates.io proxy downloads, the implementation uses a simpler ETag based on the crate name and version rather than computing a hash. This is because the crate data might not be available yet (it could be fetched from upstream), and for crates.io content, the name-version pair is sufficient as a cache key:

```rust
let etag = format!("\"{name}-{version}\"");
Ok((
    [
        (header::CONTENT_TYPE, HeaderValue::from_static("application/octet-stream")),
        (header::CACHE_CONTROL, HeaderValue::from_static("public, max-age=31536000, immutable")),
        (header::ETAG, HeaderValue::from_str(&etag).map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?),
    ],
    file,
).into_response())

```

The crates.io downloads use more aggressive caching with `public, max-age=31536000, immutable`. The max-age of 31536000 seconds is one year, which is effectively permanent. The `immutable` directive is a hint to browsers and CDNs that this content will never change, so they don't need to revalidate it even when the user does a hard refresh. This is safe for crates.io content because published crate versions are truly immutable.

The implementation changes the return type of both download functions from `Vec<u8>` to `Response` to allow attaching headers. Previously the functions returned the raw bytes and axum would automatically convert them to a response. Now the functions explicitly construct a response with headers using the tuple syntax `([(header, value)], body).into_response()`.

The PR includes tests that verify the headers are present and have the correct values:

```rust
#[tokio::test]
async fn download_returns_cache_headers() {
    // ... upload a crate ...
    
    let r = kellnr.client.oneshot(
        Request::get("/api/v1/crates/test_lib/0.2.0/download")
            .header(header::AUTHORIZATION, TOKEN)
            .body(Body::empty())
            .unwrap(),
    ).await.unwrap();

    assert_eq!(r.status(), StatusCode::OK);

    let cache_control = r.headers().get(header::CACHE_CONTROL).unwrap();
    assert_eq!(cache_control, "public, must-revalidate");

    let etag = r.headers().get(header::ETAG).unwrap().to_str().unwrap();
    assert!(etag.starts_with('"') && etag.ends_with('"'), "ETag should be quoted");
    assert_eq!(etag.len(), 66, "ETag should be a quoted SHA-256 hash (64 hex chars + 2 quotes)");
}

```

The test verifies that the ETag is properly quoted (HTTP requires quotes around ETags), and that it's exactly 66 characters (64-character SHA-256 hex digest plus two quote characters). Another test verifies that repeated downloads return the same content and headers, ensuring consistency.

The impact of this change is significant for production deployments behind a CDN or for clients making repeated requests. A cargo build that depends on 200 crates will generate 200 download requests. Without caching, every build hits the origin server. With these headers, the first build populates the CDN cache, and subsequent builds are served from cache. For a busy Kellnr instance serving many users, this can reduce origin load by 90% or more.

The different Cache-Control strategies for kellnr vs crates.io content reflect the different guarantees. Kellnr is a private registry where policies might be more flexible, so `must-revalidate` provides a safety net. Crates.io is a public registry with strong immutability guarantees, so `immutable` allows maximum caching efficiency.

### DIFF
--- a/crates/registry/src/kellnr_api.rs
+++ b/crates/registry/src/kellnr_api.rs
@@ -2659,5 +2659,4 @@ mod reg_api_tests {
                 == StatusCode::OK,
         );
     }
-
 }


crates/registry/src/kellnr_api.rs
```diff
@@ -601,11 +601,38 @@ pub async fn download(
...
         .await
     {
-        warn!("Failed to increase download counter: {e}");
+        tracing::warn!("Failed to increase download counter: {e}");
     }
 
     match cs.get(&package, &version).await {
-        Some(file) => Ok(file),
+        Some(file) => {
+            let hash = sha256::digest(&*file);
```
secana, line 609: Computing SHA-256 on every download request adds CPU overhead on the hot path. The hash is already computed at upload time in CachedCrateStorage::put() (which returns sha256::digest(&*crate_data)), but it's not stored alongside the crate data. Consider caching the hash (e.g. store it in the DB or in the cache alongside the bytes) instead of recomputing it on every download. For a large crate downloaded frequently, this is redundant work.

crates/registry/src/kellnr_api.rs
```diff
@@ -601,11 +601,38 @@ pub async fn download(
...
+            let etag = format!("\"{hash}\"");
+            Ok((
+                [
+                    (
+                        header::CONTENT_TYPE,
+                        HeaderValue::from_static("application/octet-stream"),
+                    ),
+                    (
+                        header::CACHE_CONTROL,
+                        HeaderValue::from_static("public, must-revalidate"),
```
secana, line 619: public, must-revalidate without a max-age provides almost no caching benefit. Without max-age, there's no freshness lifetime, so caches will treat the response as immediately stale. must-revalidate only
governs what happens after a cached response expires, so it has no effect when there's no expiry window to begin with.

 Additionally, the server never checks the If-None-Match request header, so even if a cache does try to revalidate using the ETag, the server will always return a full 200 response instead of 304 Not Modified. The ETag is effectively unused without server-side conditional request handling.

Either:
  - Add a max-age (e.g. public, max-age=86400, must-revalidate) and implement If-None-Match → 304 logic, or
  - Use public, max-age=31536000, immutable like the crates.io path, since kellnr also rejects re-publishing the same version (CrateExists error in storage) -> This could break if a version is deleted and then re-uploaded but that is not the norm.

crates/registry/src/kellnr_api.rs
```diff
@@ -2509,4 +2536,127 @@ mod reg_api_tests {
...
+        let etag = r.headers().get(header::ETAG).unwrap().to_str().unwrap();
+        assert!(
+            etag.starts_with('"') && etag.ends_with('"'),
+            "ETag should be quoted"
+        );
+        assert_eq!(
+            etag.len(),
+            66,
+            "ETag should be a quoted SHA-256 hash (64 hex chars + 2 quotes)"
+        );
```
secana, line 2587: This test verifies the bodies match across two downloads, but doesn't assert that the ETags are equal across calls — which is the more interesting caching property. Also, the third download inside the
assert!() at the end only checks status code and doesn't add value over the first two assertions. Consider replacing it with an ETag consistency check, and adding a test that sends If-None-Match with the ETag (once that's implemented server-side).

nyurik:
In general, etag is very helpful for any kind of **mutable** resources (per URL).  If the same URL will always produce identical result, we might not benefit from etag at all -- e.g. `GET /crate/v1.0.0/file` should be immutable, thus cache headers should just indicate a long maxage.  Note that if this file can be yanked, that makes it into a mutable resource, so etag might be more useful.

I had the same concerns as @secana about SHA256 computation. For the non-mutable but yankable resources, I would use some simple value, perhaps even a constant that depends on yanked status. For modifiable resources, just compute FNV-1a or xxHash value for speed because cryptographic certainty is not needed

secana:
Closed as it adds a lot of complexity for an unknown performance gain.

PR #1100, opened 2026-02-25
feat(perf): configurable timeouts and concurrency limits for downloads

perfectra1n:
This is a continuation of #1093, split up into 4 PRs. The requested changes from #1093 have also been implemented. This is the 4th PR.

This PR implements configurable timeouts and concurrency limits to protect Kellnr from resource exhaustion and improve reliability under load. The changes prevent scenarios where slow downloads or upstream services can tie up server resources indefinitely, and provide operators with control over how many concurrent downloads the server will handle.

The timeout configuration appears in three places reflecting three different operation types. First, there are timeouts for HTTP requests to upstream services like crates.io. These are configured in the proxy settings and apply when Kellnr fetches crate metadata or downloads from crates.io:

```rust
pub struct Proxy {
    // ... existing fields ...
    
    /// Connect timeout in seconds for upstream requests
    #[default(5)]
    pub connect_timeout_seconds: u64,

    /// Request timeout in seconds for upstream downloads
    #[default(30)]
    pub request_timeout_seconds: u64,
}

```

The connect timeout limits how long to wait for a TCP connection to establish, while the request timeout limits the entire request duration including data transfer. These are applied when building HTTP clients:

```rust
pub fn build_client(connect_timeout: Duration, request_timeout: Duration) -> Client {
    ClientBuilder::new()
        .gzip(true)
        .connect_timeout(connect_timeout)
        .timeout(request_timeout)
        .default_headers(headers)
        .build()
        .unwrap()
}

```

S3 storage gets equivalent timeout configuration because S3 operations can also hang if the S3-compatible service is slow or unresponsive:

```rust
pub struct S3 {
    // ... existing fields ...
    
    /// S3 connect timeout in seconds
    #[default(5)]
    pub connect_timeout_seconds: u64,

    /// S3 request timeout in seconds
    #[default(30)]
    pub request_timeout_seconds: u64,
}

```

These are applied to the S3 client using the object\_store library's ClientOptions:

```rust
let client_options = ClientOptions::new()
    .with_connect_timeout(Duration::from_secs(settings.s3.connect_timeout_seconds))
    .with_timeout(Duration::from_secs(settings.s3.request_timeout_seconds));
s3 = s3.with_client_options(client_options);

```

The most visible change is the per-request timeout and concurrency limit on download endpoints. These settings control how Kellnr handles incoming download requests from cargo clients:

```rust
pub struct Registry {
    // ... existing fields ...
    
    /// Download request timeout in seconds (0 = disabled)
    #[default(60)]
    pub download_timeout_seconds: u64,

    /// Max concurrent download requests (0 = unlimited)
    #[default(20)]
    pub download_max_concurrent: usize,
}

```

The concurrency limit is implemented using a tokio Semaphore shared across both kellnr and crates.io download routes. The semaphore is created at startup if the limit is non-zero:

```rust
let download_semaphore = if state.settings.registry.download_max_concurrent > 0 {
    Some(Arc::new(Semaphore::new(
        state.settings.registry.download_max_concurrent,
    )))
} else {
    None
};

```

Each download route is wrapped in middleware that acquires a permit before processing the request:

```rust
if let Some(semaphore) = download_semaphore {
    download_router = download_router.layer(middleware::from_fn(
        move |req: axum::extract::Request, next: Next| {
            let sem = semaphore.clone();
            async move {
                let _permit = sem.acquire().await.expect("download semaphore closed");
                next.run(req).await
            }
        },
    ));
}

```

When the semaphore's permit count is exhausted, additional download requests will wait until a permit becomes available. This prevents the server from accepting more downloads than it can handle, which could lead to memory exhaustion or thread pool starvation. The permit is held for the entire duration of the download, including disk or S3 I/O.

The timeout is implemented using tower-http's TimeoutLayer:

```rust
if settings.download_timeout_seconds > 0 {
    let timeout_secs = settings.download_timeout_seconds;
    download_router = download_router
        .layer(middleware::map_response(
            move |response: axum::response::Response| async move {
                if response.status() == StatusCode::GATEWAY_TIMEOUT {
                    tracing::warn!(
                        "Download request timed out after {timeout_secs}s. \
                         Consider increasing registry.download_timeout_seconds"
                    );
                }
                response
            },
        ))
        .layer(TimeoutLayer::with_status_code(
            StatusCode::GATEWAY_TIMEOUT,
            Duration::from_secs(timeout_secs),
        ));
}

```

The TimeoutLayer aborts requests that exceed the timeout and returns a 504 Gateway Timeout status. The response mapping middleware logs a warning when this happens, which helps operators tune the timeout value appropriately. The warning suggests increasing the timeout, making the system self-documenting.

These changes work together to make Kellnr resilient to various failure modes. If crates.io goes down or becomes very slow, the connect and request timeouts prevent Kellnr from hanging indefinitely waiting for upstream responses. If a cargo client starts a download and then abandons it, the download timeout will eventually clean up the server resources. If many clients request large crates simultaneously, the concurrency limit prevents the server from trying to serve them all at once and running out of memory or file handles.

The configuration defaults are conservative. A 5-second connect timeout is enough for any reasonable network, and a 30-second request timeout is sufficient for downloading most crates even on slow connections. The 60-second download timeout allows even multi-megabyte crates to download on modest connections, and the 20 concurrent download limit is reasonable for a server with a few hundred megabytes of memory. Operators can tune these values based on their deployment characteristics.

The code also handles the "feature flag" pattern cleanly. Setting a timeout to 0 disables it, setting the concurrency limit to 0 makes it unlimited. The runtime checks these values and conditionally applies the middleware layers, so there's no overhead when features are disabled. This gives operators full control without forcing them to rebuild Kellnr with different feature flags.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4015,6 +4015,7 @@ name = "kellnr-registry"
 version = "6.0.3"
 dependencies = [
  "axum",
+ "bytes",
  "chrono",
  "cookie",
  "http-body-util",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ tempfile = "3"
 testcontainers = { version = "0.27.1", features = ["http_wait_plain"] }
 thiserror = "2.0.18"
 time = "0.3.47"
-tokio = { version = "1.50.0", features = ["macros"] }
+tokio = { version = "1.50.0", features = ["macros", "signal"] }
 toml = "0.9.11"
 tower = { version = "0.5.2", features = ["util", "limit"] }
 tower-http = { version = "0.6.8", features = ["fs", "trace", "timeout"] }
--- a/crates/appstate/src/lib.rs
+++ b/crates/appstate/src/lib.rs
@@ -6,6 +6,7 @@ use flume::Sender;
 use kellnr_common::cratesio_prefetch_msg::CratesioPrefetchMsg;
 use kellnr_common::token_cache::TokenCacheManager;
 use kellnr_db::DbProvider;
+use kellnr_db::download_counter::DownloadCounter;
 use kellnr_settings::Settings;
 use kellnr_storage::cached_crate_storage::DynStorage;
 use kellnr_storage::cratesio_crate_storage::CratesIoCrateStorage;
@@ -24,6 +25,7 @@ pub type SigningKeyState = axum::extract::State<Key>;
 pub type CratesIoPrefetchSenderState = axum::extract::State<Sender<CratesioPrefetchMsg>>;
 pub type TokenCacheState = axum::extract::State<Arc<TokenCacheManager>>;
 pub type ToolchainStorageState = axum::extract::State<Option<Arc<ToolchainStorage>>>;
+pub type DownloadCounterState = axum::extract::State<Arc<DownloadCounter>>;
 
 #[derive(Clone, FromRef)]
 pub struct AppStateData {
@@ -36,10 +38,11 @@ pub struct AppStateData {
     pub cratesio_prefetch_sender: Sender<CratesioPrefetchMsg>,
     pub token_cache: Arc<TokenCacheManager>,
     pub toolchain_storage: Option<Arc<ToolchainStorage>>,
+    pub download_counter: Arc<DownloadCounter>,
 }
 
 pub fn test_state() -> AppStateData {
-    let db = Arc::new(kellnr_db::mock::MockDb::new());
+    let db: Arc<dyn DbProvider> = Arc::new(kellnr_db::mock::MockDb::new());
     let signing_key = Key::generate();
     let settings = Arc::new(kellnr_settings::test_settings());
     let kellnr_storage = Box::new(FSStorage::new(&settings.crates_path()).unwrap()) as DynStorage;
@@ -50,6 +53,7 @@ pub fn test_state() -> AppStateData {
     ));
     let (cratesio_prefetch_sender, _) = flume::unbounded();
     let token_cache = Arc::new(TokenCacheManager::new(false, 60, 1000));
+    let download_counter = Arc::new(DownloadCounter::new(db.clone(), 30));
     AppStateData {
         db,
         signing_key,
@@ -59,5 +63,6 @@ pub fn test_state() -> AppStateData {
         cratesio_prefetch_sender,
         token_cache,
         toolchain_storage: None, // Toolchain storage disabled in tests by default
+        download_counter,
     }
 }
--- a/crates/common/src/normalized_name.rs
+++ b/crates/common/src/normalized_name.rs
@@ -6,7 +6,7 @@ use crate::original_name::OriginalName;
 
 /// Index name is a lowercase version of the crate name
 
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
 pub struct NormalizedName(String);
 
 impl NormalizedName {
--- a/crates/common/src/version.rs
+++ b/crates/common/src/version.rs
@@ -1,6 +1,7 @@
 use std::cmp::Ordering;
 use std::convert::TryFrom;
 use std::fmt;
+use std::hash::{Hash, Hasher};
 use std::ops::Deref;
 
 use sea_orm::Value;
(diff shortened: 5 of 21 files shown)


crates/common/src/cratesio_downloader.rs
```diff
@@ -1,20 +1,28 @@
...
+        .connect_timeout(connect_timeout)
+        .timeout(request_timeout)
         .default_headers(headers)
         .build()
         .unwrap()
-});
+}
+
+/// Default client with sensible timeouts (5s connect, 30s request).
+pub static CLIENT: std::sync::LazyLock<Client> =
```
secana, line 24: The proxy timeout settings (proxy.connect_timeout_seconds, proxy.request_timeout_seconds) are applied to the prefetch client via get_client() in cratesio_prefetch_api.rs, but this static CLIENT — which is used by download_crate() for the actual crate download on a cache miss — still has hardcoded 5s/30s timeouts. The configured proxy timeouts don't apply to the most important path: fetching crate data from crates.io when it's not cached. download_crate() should also accept or use the configured timeouts.

crates/kellnr/src/routes/cratesio_api_routes.rs
```diff
@@ -1,19 +1,72 @@
...
 /// Creates the crates.io API routes
-pub fn create_routes(state: AppStateData) -> OpenApiRouter<AppStateData> {
+pub fn create_routes(
+    state: AppStateData,
+    download_semaphore: Option<Arc<Semaphore>>,
+) -> OpenApiRouter<AppStateData> {
+    let settings = &state.settings.registry;
+
+    // Download route with concurrency limit and timeout to prevent I/O starvation
+    let mut download_router = Router::new().route(
```
secana: This semaphore + timeout middleware setup (~30 lines) is duplicated verbatim in kellnr_api_routes.rs:37. Extract it into a shared helper function, e.g. in routes/mod.rs:

```rust
  fn apply_download_limits(
      router: Router<AppStateData>,
      semaphore: Option<Arc<Semaphore>>,
      settings: &Registry,
  ) -> Router<AppStateData>
```

crates/kellnr/src/routes/cratesio_api_routes.rs
```diff
@@ -1,19 +1,72 @@
...
+                    if response.status() == StatusCode::GATEWAY_TIMEOUT {
+                        tracing::warn!(
+                            "Download request timed out after {timeout_secs}s. \
+                             Consider increasing registry.download_timeout_seconds"
+                        );
+                    }
+                    response
+                },
+            ))
+            .layer(TimeoutLayer::with_status_code(
```
secana: Due to axum's layer ordering (last added = outermost), the TimeoutLayer wraps the semaphore middleware. This means the 60s timeout clock starts before a semaphore permit is acquired. If the server is at capacity, a request could spend most of its timeout budget waiting in the semaphore queue, leaving little time for the actual download. This may be intentional (total request time limit), but if the intent is to timeout the download itself, the layer order should be reversed (add timeout before semaphore). Either way, worth a comment explaining the chosen ordering.

crates/kellnr/src/routes/cratesio_api_routes.rs
```diff
@@ -1,19 +1,72 @@
...
+        "/dl/{package}/{version}/download",
+        get(cratesio_api::download),
+    );
+
+    if let Some(semaphore) = download_semaphore {
+        download_router = download_router.layer(middleware::from_fn(
+            move |req: axum::extract::Request, next: Next| {
+                let sem = semaphore.clone();
+                async move {
+                    let _permit = sem.acquire().await.expect("download semaphore closed");
```
secana: When the semaphore is full, requests block silently until a permit is available (or the outer timeout fires with a generic 504). Consider using try_acquire() and returning 503 Service Unavailable or 429. Too Many Requests immediately when the server is at capacity. This gives clients useful feedback (they can retry with backoff) instead of queuing silently until timeout. Cargo handles non-200 responses gracefully already.

secana:
For all PR: If they are merged and the PR introduces new config values, please add them to the docs here https://github.com/kellnr/website and the helm chart here https://github.com/kellnr/helm

secana:
Default for any new value should not conflict with the current behavior, e.g. should be unlimited or disabled and only be set manually by the user if needed to not change the behavior of a running instance.

PR #1106, opened 2026-02-27
Enable support for splitting storage backends over registry-owned crates, proxied crates, and toolchain data.

aalekhpatel07:
Consider a scenario where a S3 service is running on HDDs, but the server infra is running on SSDs. Here it might make sense to have a split storage, i.e. the registry-owned crates would be persisted in a robust (but possibly slow) s3-compatible storage while any proxied crates can be cached locally on the same filesystem as the server for quick downloads.

I happen to run into this hardware configuration, so I took a chance at implementing this split storage configuration.

I'm open to ideas/suggestions that would improve this PR and hopefully be a valuable feature addition to Kellnr.

Thanks for all your amazing work in building and maintaining Kellnr.

### DIFF
--- a/crates/kellnr/src/main.rs
+++ b/crates/kellnr/src/main.rs
@@ -14,7 +14,7 @@ use kellnr_db::{ConString, Database, DbProvider, PgConString, SqliteConString};
 use kellnr_index::cratesio_prefetch_api::{
     CratesIoPrefetchArgs, UPDATE_CACHE_TIMEOUT_SECS, init_cratesio_prefetch_thread,
 };
-use kellnr_settings::{CliResult, LogFormat, Settings, ShowConfigOptions, parse_cli};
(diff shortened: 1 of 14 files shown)


secana:
Hi @aalekhpatel07,

thank you for taking the time to implement this and for the kind words about Kellnr. I really appreciate the effort you put into the PR.

After thinking about it for a while, I decided not to merge it. Splitting the storage backends between registry-owned crates, proxied crates, and toolchain data adds a fair amount of complexity to the configuration and the storage layer, and it is a setup that only very few users will need. Every additional configuration permutation is something I have to test, document, and maintain long-term, so I try to be careful about which ones I take on.

There is also a mitigating factor for the performance concern behind this: Kellnr caches crates heavily in RAM, so after the first access most downloads are served from memory and the slow HDD-backed storage is rarely on the hot path.

I'm sorry to turn down a working contribution, and I hope you understand the reasoning. If the caching turns out not to be enough for your setup, I'd be happy to hear about it, maybe there is a simpler way to address it.

Thanks again for contributing!

PR #1135, opened 2026-03-17
fix oath autocreate username problem

### LINKED ISSUE #1134: oauth autocreate user does not use the preferred_username from the token
Hey, the automatic creation of the username via OAuth seems to be having an issue. The token correctly shows `preferred_username=“firstname.lastname”`,

but the username is created as `firstname_lastnam

MR is created

badsmoke:
fix #1134

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,7 @@
           version = "0.1.0";
           src = ./ui;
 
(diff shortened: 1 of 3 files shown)


ui/package-lock.json
secana, line 1: Please revert if it is not related to the change in the PR.

ui/package.json
secana, line 1: Please revert if it is not related to the change in the PR.

flake.nix
secana, line 1: Please revert if it is not related to the change in the PR.

PR #1235, opened 2026-06-08
fix: improve get_crate_data performance by grouping dep desc queries

Celsiusss:
`get_crate_data` loops over every crate meta then every dependency of that, then performs a query for every dependency, causing massive slowdowns when a crate has many dependencies and versions.

This (partially) solves that by grouping queries over `IndexDep`s. It still makes a query for every published crate version, but still provides a big improvement in loading times for crates with many dependencies.

Ideally the `/crate_data` endpoint should return a less verbose dataset, but that would be a breaking API change.

### DIFF
--- a/crates/db/src/database/operations.rs
+++ b/crates/db/src/database/operations.rs
@@ -23,27 +23,13 @@ use crate::error::DbError;
 use crate::password::{hash_pwd, hash_token};
 use crate::provider::DbResult;
 
-pub async fn get_desc_for_deps<'a, C, D>(
-    db_con: &C,
-    deps: D,
-) -> DbResult<HashMap<&'a IndexDep, String>>
+pub async fn get_desc_for_deps<'a, C, D>(db_con: &C, deps: D) -> DbResult<HashMap<String, String>>
 where
     C: ConnectionTrait,
     D: Iterator<Item = &'a IndexDep>,
 {
-    fn descriptions_by_dep<'a>(
-        models: impl IntoIterator<Item = (String, Option<String>)>,
-        deps: &[&'a IndexDep],
-    ) -> impl Iterator<Item = (&'a IndexDep, String)> {
-        models.into_iter().filter_map(move |(name, description)| {
-            let description = description?;
-            let dep = *deps.iter().find(|d| d.name == name)?;
-            Some((dep, description))
-        })
-    }
-
-    let mut crates = Vec::new();
     let mut cratesio_crates = Vec::new();
+    let mut crates = Vec::new();
 
     for dep in deps {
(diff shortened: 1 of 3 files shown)


crates/db/src/database/operations.rs
```diff
@@ -21,28 +23,57 @@ use crate::error::DbError;
...
+    D: Iterator<Item = &'a IndexDep>,
+{
+    fn descriptions_by_dep<'a>(
+        models: impl IntoIterator<Item = (String, Option<String>)>,
+        deps: &[&'a IndexDep],
+    ) -> impl Iterator<Item = (&'a IndexDep, String)> {
+        models.into_iter().filter_map(move |(name, description)| {
+            let description = description?;
+            let dep = *deps.iter().find(|d| d.name == name)?;
+            Some((dep, description))
```
secana: Thanks for tackling this, the batching is a great improvement! I think there's one edge case worth handling: this mapping can drop descriptions for duplicate dependency names within a single version.

The DB query returns one row per unique crate name, and `deps.iter().find(...)` matches only the first `IndexDep` with that name, so any further dep sharing the name never gets an entry. As a result `dep_descs.get(&dep)` returns `None` and that dependency ends up with no description.

It's a fairly common case, since a crate often lists the same dependency name more than once (normal plus dev/build, or target-specific entries like `[target.'cfg(unix)'.dependencies] libc` alongside `[target.'cfg(windows)'.dependencies] libc`). The previous per-dep code returned a description for each of them. There's no test covering this, which is probably why CI stays green.

One option would be to key the lookup by name instead of by `&IndexDep`, then map every dep. That fixes the case above and has the nice side effect of removing the need for the new `Hash` derives on `IndexDep` and `DependencyKind`:

```rust
// name -> description
let cratesio_desc: HashMap<String, String> = cratesio_models
    .into_iter()
    .filter_map(|m| Some((m.name, m.description?)))
    .collect();
let local_desc: HashMap<String, String> = models
    .into_iter()
    .filter_map(|m| Some((m.name, m.description?)))
    .collect();
```

Then in `get_crate_data` you could look up each dep by `dep.name`, picking the map based on the registry (crates.io vs local), so all deps sharing a name get the description. Happy to help if useful!

secana (commented):
Nice improvement, the per-dependency N+1 query was a real problem and this is a solid fix. I left one note inline about an edge case to handle before merge. Otherwise the approach and scope look great.

crates/db/src/database/operations.rs
Celsiusss: Good catch! This also simplified the code a bit

PR #1275, opened 2026-07-07
feat(auth): enforce SSO-only login with RP-initiated logout

hwittenborn:
## Summary

Adds two new OAuth2/OIDC settings that let an operator make SSO the sole authentication path:

- `oauth2.enforced` — disables local password login, change, reset, and user creation, so the only way in is via the SSO provider.
- `oauth2.auto_redirect` — skips the Kellnr login page entirely and sends the browser straight to the provider (requires `enforced`).

It also wires up RP-Initiated Logout so signing out of Kellnr can end the upstream provider session too.

## Changes

- Settings: add `enforced` / `auto_redirect` flags with `validate()` guards (`enforced` requires `enabled`; `auto_redirect` requires `enforced`), validated at startup in `main.rs`. Extract origin URL building into `Origin::base_url()`, replacing the ad-hoc callback-URL construction.
- Backend enforcement: `ensure_local_auth_enabled` rejects the local password routes (`login`, `add`, `reset_pwd`, `change_pwd`) with 403 when SSO is enforced.
- RP-initiated logout: read the provider's `end_session_endpoint` from the existing discovery fetch (via a custom `RpLogoutMetadata` provider metadata type — no extra round-trip), store the raw ID token in an encrypted cookie at callback, and have `logout` return an `end_session_url` for the UI to redirect to.
- Frontend: expose the new config, hide the password form and redirect from the login page when enforced/auto_redirect, follow the logout URL, and cache the OAuth2 config lookup.

## Test

- New `validate()` unit tests covering the enforced/auto_redirect dependency rules.
- Existing settings and auth suites still pass.

### DIFF
--- a/crates/auth/src/oauth2.rs
+++ b/crates/auth/src/oauth2.rs
@@ -10,13 +10,13 @@ use kellnr_settings::OAuth2 as OAuth2Settings;
 use openidconnect::core::{
     CoreAuthDisplay, CoreAuthPrompt, CoreAuthenticationFlow, CoreClient, CoreErrorResponseType,
     CoreGenderClaim, CoreIdTokenClaims, CoreJsonWebKey, CoreJweContentEncryptionAlgorithm,
-    CoreProviderMetadata, CoreRevocationErrorResponse, CoreTokenIntrospectionResponse,
-    CoreTokenResponse,
+    CoreRevocationErrorResponse, CoreTokenIntrospectionResponse, CoreTokenResponse,
 };
 use openidconnect::{
     AuthorizationCode, ClaimsVerificationError, ClientId, ClientSecret, CsrfToken,
     EmptyAdditionalClaims, EndpointMaybeSet, EndpointNotSet, EndpointSet, IssuerUrl, Nonce,
-    PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, Scope, TokenResponse, reqwest,
+    PkceCodeChallenge, PkceCodeVerifier, ProviderMetadataWithLogout, RedirectUrl, Scope,
+    TokenResponse, reqwest,
 };
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -109,6 +109,8 @@ pub struct TokenResult {
     pub claims: CoreIdTokenClaims,
     /// Raw JWT payload for extracting additional claims
     pub raw_payload: serde_json::Value,
+    /// Raw ID token JWT, used as `id_token_hint` for RP-initiated logout
+    pub id_token: String,
 }
 
 /// OAuth2/OIDC authentication handler
@@ -122,6 +124,7 @@ pub struct OAuth2Handler {
     settings: Arc<OAuth2Settings>,
     issuer_url: IssuerUrl,
     http_client: reqwest::Client,
+    end_session_endpoint: Option<String>,
 }
 
 impl OAuth2Handler {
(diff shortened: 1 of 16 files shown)


secana (changes requested):
Hi @hwittenborn, thanks a lot for this contribution. SSO-only enforcement and RP-initiated logout are features I'd really like to have in Kellnr, and the overall structure (settings validation, reusing the discovery fetch for `end_session_endpoint`, the `Origin::base_url()` refactor) is well done.

That said, I found a few issues during review that I'd consider blockers before merging:

1. **Enforcement doesn't cover the registry API.** `crates/auth/src/token.rs` still authenticates `Basic user:password` against local password hashes, so with `oauth2.enforced = true` a user deprovisioned at the IdP can keep pushing/pulling crates. And since `reset_pwd`/`change_pwd` now return 403, an admin can't even rotate that password anymore. I think the gate belongs in the shared auth path rather than in the four individual handlers.

2. **Total lockout when the IdP is unreachable at startup.** `init_oauth2_handler` only logs a warning and returns `None` on discovery failure. SSO login then 404s (no retry), while `enforced` still 403s all password logins. Nobody, including the admin, can log in until the IdP is back *and* Kellnr is restarted. With `enforced = true` this should probably fail startup hard (like the new `validate()` call) or retry discovery.

3. **Logout can become impossible with `auto_redirect`.** If `/logout` returns no `logout_url` (provider without `end_session_endpoint`, or missing ID-token cookie), the UI does `router.push('/')`, the auth guard sends the user to the login page, auto-redirect forwards to the IdP, and the still-active IdP session silently logs them back in. There's no way to end the session.

4. **The `!('error' in to.query)` escape hatch never fires.** Callback failures (e.g. unprovisioned user with `auto_provision_users = false`) return bare 403/400 responses, and the `/api/v1/oauth2/error` route isn't the registered redirect URI. So with `auto_redirect`, users get trapped bouncing between the IdP and a bare error page.

Smaller points:

- Storing the raw ID token in a cookie can exceed the 4 KB browser limit for IdPs with large group claims (encryption adds ~37%), silently breaking RP-logout for exactly those setups. Since `id_token_hint` is only RECOMMENDED by the spec, a fallback without it, or storing the token server-side with the session, would be more robust. Related: the JWKS-refresh path discards the re-fetched `end_session_endpoint`.
- The admin UI still shows Add User / Reset Password / Change Password under `enforced`, and the resulting 403 redirects the admin to the login page (`shouldRedirectToLogin` treats 403 like 401).
- The `ui/package-lock.json` diff looks like unrelated npm-version churn, please revert it.
- It would be great to cover the new login/logout behaviors in the Playwright suite; `tests/src/ui-oauth2.spec.ts` with the mock OIDC server (which already advertises an `end_session_endpoint`) should make that straightforward.

Happy to merge once the four blockers are addressed. Thanks again for the work you put into this!

secana:
@hwittenborn Hi! Are you still interested in moving the draft forward?

hwittenborn:
Hey @secana, apologies for the delay on this... I have been very busy at my day job and completely lost track of this.

I do not think I have the time to dig deeper into this unfortunately, though if you think this is valuable, feel free to continue it or re-assign it. You can also close it for now, and I would be more then fine opening a new MR if/when I get time for it.

aergus-ferrous:
Hello, my employer @ferrous-systems is interested in finishing the work on this PR. I am working on addressing the blockers.

PR #1349, opened 2026-08-27
feat(storage): add Google Cloud Storage backend support

aarnaud:
Adds GCS as a third storage backend alongside local filesystem and S3, for GCP-hosted deployments. Auth is Application Default Credentials only, matching how kellnr never manages OS-level credentials for other backends. Includes Rust integration tests against fake-gcs-server and Playwright e2e specs mirroring the existing S3 coverage.


secana:
Thanks for the great PR. Very welcome addition. The feature will be in the next release.

PR #1357, opened 2026-08-30
fix(auth): allow configurable additional audiences

moderny-it:
Hey, while trying out kellnr with my IdP (Zitadel) I saw it sending multiple aud in the token during oauth2 and kellnr right now need the aud to only have the clientId in aud.

With the related PR for the helm chart and this PR, it allows it to accept "extra" aud.

kellnr helm PR : https://github.com/kellnr/helm/pull/93

### DIFF
--- a/crates/auth/src/oauth2.rs
+++ b/crates/auth/src/oauth2.rs
@@ -306,8 +306,9 @@ impl OAuth2Handler {
 
         let nonce = Nonce::new(nonce.to_string());
 
-        // An ID token may contain multiple audiences. When configured, accept
-        // additional audiences while requiring the configured client ID.
+        // An ID token may contain multiple audiences. Require the configured
+        // client ID and accept additional audiences only when explicitly
+        // allowlisted.
         let verifier = client.id_token_verifier().set_other_audience_verifier_fn({
             let additional_audiences = self.settings.additional_audiences.clone();
             move |aud| {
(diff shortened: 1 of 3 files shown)


secana:
Hi, thanks a lot for this PR!

Three things I'd like to change before merging:

1. An allowlist instead of a boolean

Right now `move |_| true` trusts any audience value at all. The reason that hook takes a closure rather than a flag is so you can name the audiences you trust and in your case you do know them, they're your Zitadel client IDs. Something like:

```rust
/// Additional audiences accepted in the ID token besides the client ID
#[configurable(env_list)]
#[arg(value_delimiter = ',')]
pub additional_audiences: Vec<String>,

let extra = self.settings.additional_audiences.clone();
let verifier = client.id_token_verifier()
    .set_other_audience_verifier_fn(move |aud| extra.iter().any(|a| a == aud.as_str()));
```

2. Keep coverage for the single-audience case

`include_client_audience` defaults to `true`, which means `sign_id_token` now always emits `["other-audience", "test-client"]`. That leaves no test anywhere exercising a plain single-audience token, which is what the default configuration sees in production. Could you either flip the default so the baseline stays single-audience, or add a case asserting a strict handler accepts `aud == [client_id]`?

3. Assert on the actual error

Both new `assert!(...is_err())` checks pass on any failure, so a nonce mismatch or an expired token would satisfy them just as well as the `InvalidAudience` you're aiming at. Matching the variant (or checking the message contains audience) would make them prove what the comments say they prove.

Happy to merge after that.

moderny-it:
Sounds good, will do that, fit the helm to the changes, test on my cluster and if everything works will update the PR :)

moderny-it:
It is done, we are now checking for the additional audience values instead of the boolean.

The test exchange_recovers_from_signing_key_rotation test the clientId audience only at first and then we test when there is only an additional audience to fail, then we test clientId + additional to work and finally when the token has clientId + additional but no additional configured so it fails

secana:
Thanks! Looks good.

Issue #1348, opened 2026-08-27
Feature proposal: search by name and description


hishamhm:
Hi! Thank you for Kellnr! Following the recommendation that I saw in CONTRIBUTING.md, I'm opening an issue to discuss a new feature before opening a PR.

This is a feature proposal to add support for searching by name and description, using the Registry Web API search endpoint (`/api/v1/crates?q=`).

> What is its use case?

I need to search for crates by some extra data other than its name. Since crates.io supports searching by description, I figured that adding this extra data to the `description` field of a crate allowed for finding them without having to extend the search API. That works well for crates.io, but not for Kellnr, which searches only by name.

> Why should it be in Kellnr?

To better match the Registry Web API implementation as done by crates.io, thus improving compatibility with it.

> Any relevant technical details

By the letter of the spec, the Registry Web API [does not specify the search criteria](https://doc.rust-lang.org/cargo/reference/registry-web-api.html#search): "The search request will perform a search for crates, using criteria defined on the server", but crates.io does perform search based on description contents. See, e.g., `curl -v 'https://crates.io/api/v1/crates?q=vcard%20parser' -H 'user-agent:testing' | jq .`

Description search in crates.io is case-insensitive. I do not need this functionality myself, but if necessary to adopt the feature, we could either do it "portably" via the sea-orm equivalent of `LOWER(description) LIKE LOWER(pattern)`, or adopting the specific functionality for each backend:
* SQLite: apparently `LIKE` is [already case-insensitive for ASCII characters in SQLite](https://stackoverflow.com/questions/15480319/case-sensitive-and-insensitive-like-in-sqlite#comment38404510_15480401). 
* Postgres: it [has ILIKE](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE) and sea-orm supports this extension; [it can](https://dba.stackexchange.com/a/156974) be configured with case insensitive collation in the `description` field, but that would require a migration and seems more trouble than it's worth.
* MySQL: the case-insensitive collation can be [applied in the query](https://stackoverflow.com/a/18853475/1793220).

I already have a prototype of search by name and description written (modulo case-insensitivity). Happy to open a PR if the feature is welcome!

secana:
This was implemented by @hishamhm in #1358 (thanks again for the contribution!) and merged to main, including a follow-up hardening fix in #1365. Search on the registry API endpoint now matches case-insensitively against both crate name and description. The feature will be part of the next release, so I'm closing this as completed.

PR #1358, opened 2026-09-01
feat(api): search by crate name and description

hishamhm:
As described in https://github.com/kellnr/kellnr/issues/1348.

This adds support for searching by name and description, using the Registry Web API search endpoint (`/api/v1/crates?q=`). 

The Registry Web API [does not specify the search criteria](https://doc.rust-lang.org/cargo/reference/registry-web-api.html#search): "The search request will perform a search for crates, using criteria defined on the server", but crates.io does perform search based on description contents. See, e.g., `curl -v 'https://crates.io/api/v1/crates?q=vcard%20parser' -H 'user-agent:testing' | jq .`. This PR should reduce the feature gap with the Registry Web API implementation from crates.io.


secana:
Hi, thanks for this!

Two things I'd like to sort out before merging:

1. The web UI still rejects the queries this enables. `crates/web-ui/src/ui.rs` keeps name: `OriginalName`, so only the registry API side was relaxed. I hit the UI endpoint directly:

`GET /search?name=vcard%20parser  ->  400 Bad Request`

Could you give the UI param the same treatment as `search_params.rs`? Its `/// Search` for crates by name doc comment needs updating too.

2. Description matching is case-sensitive on PostgreSQL. I ran a `#[db_test]` (which generates both a sqlite and a postgres variant) with the description "A Fast VCard Parser":

| Backend | `q=vcard` | `q=VCard` |
|---|---|---|
| SQLite | 1 hit | 1 hit |
| PostgreSQL | **0 hits** | 1 hit |

This existed for names already, but it was masked because krate.name stores the lowercased normalized name. Descriptions are prose, so capitals are normal and this will bite often and crates.io search is case-insensitive. `ILIKE` on Postgres or `lower()` on both sides would do it; a test with mixed case would keep it from coming back.

Two smaller things while you're in there:

- The crates.io cache branch in `query_crates` still filters on name only, so with `cache=true` local crates match name-or-description while cached crates match name only. `CratesIoIden::Description` is already selected a few lines above.
- `q` no longer has a length bound now that `OriginalName`'s 64-char cap is gone. A cap in `SearchParams::from_request_parts` would stop a huge `q` becoming a huge `LIKE` pattern.

Thanks again, this is a good addition.

hishamhm:
@secana Thank you for the review!

I pushed an update which addresses the outstanding bits:

1. Both the UI and Registry APIs now now go through `NameOrDescription`, which should behave the same.
2. I added the `lower` call to the query so it behaves the same on all DBs, and a test demonstrating case insensitivity.
3. In `query_crates` now both `query` and `query2` now apply the same filtering expression.
4. `NameOrDescription` enforces a 64-character length bound.

secana:
Thanks! The feature fill be in the next release.

Issue #771, opened 2025-08-15
config: required_crate_fields does not support inclusion of "rust-version"


sorenrade:
Adding 
```rust
pub struct PublishMetadata {
    // ...
    #[serde(rename(serialize = "rust-version"))]
    pub rust_version: Option<String>,
 }
``` 
seems to fix. The `serde(rename(serialize()))` directive is required because the checking is done [by reserializing to JSON](https://github.com/kellnr/kellnr/blob/9612b253a5e2fdc36c13085cb4e190479a7ccab0/crates/registry/src/kellnr_api.rs#L390) and then checking if the fields are set. Maybe a cleaner solution is to carry a `vec` of all the TOML strings through from where they are initially parsed, so this does not need to be maintained? 

Thanks to all maintainers who contributed to a great project!

secana:
Thanks for reporting the issue. A fix will be in the next release.

PR #1363, opened 2026-09-03
fix(index): keep rust_version in served crates.io index entries

### LINKED ISSUE #1362: crates.io proxy drops rust_version from served index entries, breaking MSRV-aware resolution
### Summary

Kellnr's crates.io proxy drops the `rust_version` field from the sparse index entries it serves. Cargo's MSRV-aware resolver, which is the default for edition 2024 / `resolver = "3"`, therefore sees every proxied crate as having no minimum supported Rust version and can select releases the active toolchain cannot compile.

### Reproduction

Against Kellnr 6.8.0 with `proxy.enabled = true`, comparing `GET /api/v1/cratesio/{prefix}/{crate}` with the same path on `index.crates.io`:

| crate | versions (upstream / Kellnr) | entries with `rust_version` (upstream / Kellnr) |
| --- | --- | --- |
| anyhow | 106 / 106 | 60 / 0 |
| clap | 461 / 461 | 197 / 0 |
| serde | 316 / 316 | 100 / 0 |
| tokio | 195 / 195 | 96 / 0 |
| regex | 169 / 169 | 30 / 0 |
| reqwest | 125 / 125 | 47 / 0 |
| serde_json | 183 / 183 | 83 / 0 |
| clap_builder | 108 / 108 | 108 / 0 |

The version set, `cksum` and `yanked` all match exactly. Only `rust_version` is missing, on 100% of entries.

### Cause

Proxied index lines are not relayed verbatim. They are deserialized into `IndexMetadata` (`crates/common/src/index_metadata.rs`), persisted, and re-serialized from database columns (`crates/db/src/database/operations.rs`). `IndexMetadata` has no `rust_version` field, so the value is discarded on the way in.

Note that `PublishMetadata` does carry `rust_version` (added for #771), but that is a separate struct used for the publish-time field check, so the value never reaches the served index.

### Why it is easy to miss

Every integrity check still passes: checksums match, `Cargo.lock` verifies, the version set is complete. Only the resolver's input is degraded, and the symptom surfaces later as a compile error in a transitive dependency the user never selected.

### Proposed fix

Add `rust_version: Option<String>` to `IndexMetadata` (the index spells it `rust_version`, unlike the `rust-version` manifest key), persist it in `crate_index` and `cratesio_index` via a migration, and populate it on both the proxy path and Kellnr's own publish path. This mirrors how the `pubtime` index field was added in #970.

I have this implemented and tested locally and will open a PR referencing this issue.

---

*Disclosure: this issue and the accompanying patch were produced with AI assistance (Claude Code). The reproduction numbers above are from real runs against crates.io and a local Kellnr 6.8.0 instance, and the change has been built and tested; a human reviewed both before submission.*

seemethere:
Fixes #1362.

The crates.io proxy drops `rust_version` from the sparse index entries it serves, so cargo's MSRV-aware resolver (the default for edition 2024 / `resolver = "3"`) sees no minimum supported Rust version for any proxied crate and can select releases the active toolchain cannot compile. Every integrity check still passes — checksums, `yanked`, the full version set — which makes it easy to miss; the symptom appears later as a compile error in a transitive dependency the user never chose.

Proxied lines are not relayed verbatim: they are deserialized into `IndexMetadata`, persisted, and re-serialized from database columns. `IndexMetadata` had no `rust_version` field, so the value was discarded on the way in.

### Changes

1. **`fix(index): keep rust_version in served index entries`** — adds `rust_version: Option<String>` to `IndexMetadata` (`#[serde(default, skip_serializing_if = "Option::is_none")]`; the index spells it `rust_version`, unlike the `rust-version` manifest key), a nullable TEXT column on `crate_index` and `cratesio_index` via migration `m20260903_000001_index_rust_version`, and wiring on the read and write paths. Kellnr's own publish path is covered too — the value comes from `PublishMetadata`, which already carried it since #771.
2. **`fix(index): refresh rust_version on already-cached versions`** — `add_cratesio_prefetch_data` previously refreshed only `yanked` for a version it already had, so after upgrading, existing cache rows would keep a NULL MSRV forever. The existing-row branch now also refreshes `rust_version`.

Structured after #970, which added the `pubtime` index field; the migration uses the `iden.rs` enums like the most recent migrations rather than the older snapshot-entities style.

### Verification

Measured against `index.crates.io` with a local 6.8.0 instance, before and after. Version sets and `cksum`/`yanked` match in every case; only the MSRV column changes:

| crate | versions | `rust_version` upstream | before | after |
| --- | --- | --- | --- | --- |
| anyhow | 106 | 60 | 0 | 60 |
| clap | 461 | 197 | 0 | 197 |
| serde | 316 | 100 | 0 | 100 |
| tokio | 195 | 96 | 0 | 96 |
| regex | 169 | 30 | 0 | 30 |
| reqwest | 125 | 47 | 0 | 47 |
| serde_json | 183 | 83 | 0 | 83 |
| clap_builder | 108 | 108 | 0 | 108 |

Tests added: `rust_version_round_trips_losslessly` and `rust_version_none_is_omitted_from_serialization` (`kellnr-common`); `cratesio_index_keeps_rust_version`, `crate_index_keeps_rust_version` and `cratesio_index_heals_missing_rust_version` (`kellnr-db`). `cargo fmt --all -- --check` and `cargo check --workspace --all-targets` are clean; `cargo test -p kellnr-db` sqlite suite is 81/81. The `postgres_*` variants could not run here (testcontainers wants a Docker socket; this box has podman) — that applies to the existing postgres tests too. The migration is plain `ALTER TABLE ... ADD COLUMN <nullable text>`, portable across both backends.

### Upgrade behaviour, worth a maintainer's opinion

An existing cache heals only when crates.io answers a refresh with 200. The background update sends the stored ETag, so unchanged crates get a 304 and `add_cratesio_prefetch_data` is never reached — in practice a crate backfills the first time a new version is published upstream after the upgrade. Busy crates recover within hours; dormant ones may not. Verified: over a 617-row cache built by unpatched 6.8.0, anyhow went 0/106 → 60/106 once a 200 was forced, while serde and tokio stayed at 0 behind their 304s. Operators wanting an immediate full backfill can blank `cratesio_crate.e_tag`/`last_modified` and restart (~15s for anyhow here).

Happy to add a one-shot backfill to the migration, or an "ignore stored validators once after upgrade" path, if you would prefer the upgrade to be self-healing — I left it out to keep this change small.

Related, **not changed here**: on the crates.io side `deps`, `features`, `features2`, `links`, `cksum`, `pubtime` and `v` are likewise insert-only, so a pre-existing cache cannot pick up upstream corrections to those either. Only `yanked` was ever refreshed. That looked like a wider pre-existing gap than this fix should take on — let me know if you would like it addressed here or in a follow-up.

---

*Disclosure: this change was produced with AI assistance (Claude Code). All measurements above are from real runs against crates.io and a local Kellnr instance, and a human reviewed the change before submission.*

### DIFF
--- a/crates/common/src/index_metadata.rs
+++ b/crates/common/src/index_metadata.rs
@@ -120,6 +120,13 @@ pub struct IndexMetadata {
     // }
     #[serde(skip_serializing_if = "Option::is_none")]
     pub features2: Option<BTreeMap<String, Vec<String>>>,
+    // The minimal supported Rust version of the package, or null if not
+    // specified. This field is optional and defaults to null.
+    //
+    // Note that the index spells this field `rust_version`, while the
+    // manifest key it originates from is `rust-version`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rust_version: Option<String>,
 }
 
 impl IndexMetadata {
@@ -197,6 +204,7 @@ impl IndexMetadata {
             links: registry_metadata.links.clone(),
             v: Some(1),
             features2: None,
+            rust_version: registry_metadata.rust_version.clone(),
         }
     }
 
@@ -212,6 +220,7 @@ impl IndexMetadata {
             pubtime: None,
             v: Some(1),
             features2: None,
+            rust_version: None,
         }
     }
 
@@ -475,6 +484,7 @@ mod tests {
             pubtime: Some(pubtime),
             v: Some(1),
             features2: None,
+            rust_version: None,
         };
 
         let json = metadata.to_json().unwrap();
@@ -499,6 +509,7 @@ mod tests {
             pubtime: None,
             v: Some(1),
             features2: None,
+            rust_version: None,
         };
 
         let json = metadata.to_json().unwrap();
@@ -547,6 +558,33 @@ mod tests {
     }
 
     #[test]
+    fn rust_version_round_trips_losslessly() {
+        // The index spells the MSRV field `rust_version`, unlike the
+        // `rust-version` manifest key it is derived from.
+        let json = r#"{"name":"anyhow","vers":"1.0.100","deps":[],"cksum":"abc","features":{},"yanked":false,"v":2,"rust_version":"1.39"}"#;
+
+        let metadata: IndexMetadata = serde_json::from_str(json).unwrap();
+
+        assert_eq!(metadata.rust_version, Some("1.39".to_string()));
+        assert_eq!(
+            serde_json::from_str::<serde_json::Value>(&metadata.to_json().unwrap()).unwrap(),
+            serde_json::from_str::<serde_json::Value>(json).unwrap()
+        );
+    }
+
+    #[test]
+    fn rust_version_none_is_omitted_from_serialization() {
+        let metadata = IndexMetadata::minimal("test", "1.0.0", "abc123");
+
+        let json = metadata.to_json().unwrap();
+
+        assert!(
+            !json.contains("rust_version"),
+            "Expected rust_version to be omitted when None, got: {json}"
+        );
+    }
+
+    #[test]
     fn serialize_indices_ends_with_newline_and_one_line_per_entry() {
         let mk = |vers: &str| IndexMetadata {
             name: "crate".to_string(),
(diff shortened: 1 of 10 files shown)


secana:
Great catch! The fix will be in the next release.

PR #1364, opened 2026-09-04
feat(docs): serve generated docs from the pluggable Storage backend

aarnaud:
Docs (rustdoc) were hardcoded to local disk via ServeDir, so deployments without persistent local storage (e.g. GCS-backed) lost generated docs and served 404s. Thread doc generation, manual publishing, and serving through the same Storage trait already used for crates/cratesio/toolchains, via a new DocsStorage wrapper with overwrite-on-publish semantics.

Bundled fix: a corrupt/path-traversal doc archive's extraction error was silently discarded, always returning 200; it now surfaces as 400.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4228,6 +4228,7 @@ dependencies = [
  "kellnr-storage",
  "kellnr-web-ui",
  "kellnr-webhooks",
+ "mime_guess",
  "mockall",
  "moka",
  "openssl",
@@ -4348,6 +4349,7 @@ name = "kellnr-docs"
 version = "6.8.0"
 dependencies = [
  "axum",
+ "bytes",
  "cargo",
  "flate2",
  "fs_extra",
@@ -4361,6 +4363,7 @@ dependencies = [
  "kellnr-registry",
  "kellnr-settings",
  "kellnr-storage",
+ "mockall",
  "serde",
  "serde_json",
  "tar",
@@ -4516,6 +4519,8 @@ version = "6.8.0"
 dependencies = [
  "async-trait",
  "bytes",
+ "chrono",
+ "futures-util",
  "kellnr-common",
  "kellnr-fakegcs-testcontainer",
  "kellnr-rustfs-testcontainer",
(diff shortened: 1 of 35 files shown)


secana:
Thanks for the PR!

A few things I'd love to see addressed before merge:

1. New required bucket: `docs_storage` is initialized unconditionally (not gated by `docs.enabled`), so S3/GCS deployments now need a kellnr-docs bucket to exist on upgrade. Could you confirm what happens on startup/first-request if that bucket is missing (hard crash vs. graceful error)? The PR looks like a breaking change. Maybe we find a solution to make it non-breaking.
2. No data migration: docs previously always lived on local disk, so existing S3/GCS users will have an empty docs bucket after upgrading until crates are republished/regenerated. We should think of a migration path.
3. Caching headers regression: the new `serve_doc_file` handler drops `ETag/Last-Modified/conditional-GET` support that `ServeDir` gave us for free, so every doc asset becomes a full re-fetch on every page view. Given docs can now be S3/GCS-backed, that's real egress cost. Can we add at least ETag/If-None-Match handling back?

aarnaud:
> Thanks for the PR!
> 
> A few things I'd love to see addressed before merge:
> 
>     1. New required bucket: `docs_storage` is initialized unconditionally (not gated by `docs.enabled`), so S3/GCS deployments now need a kellnr-docs bucket to exist on upgrade. Could you confirm what happens on startup/first-request if that bucket is missing (hard crash vs. graceful error)? The PR looks like a breaking change. Maybe we find a solution to make it non-breaking.
> 
>     2. No data migration: docs previously always lived on local disk, so existing S3/GCS users will have an empty docs bucket after upgrading until crates are republished/regenerated. We should think of a migration path.
> 
>     3. Caching headers regression: the new `serve_doc_file` handler drops `ETag/Last-Modified/conditional-GET` support that `ServeDir` gave us for free, so every doc asset becomes a full re-fetch on every page view. Given docs can now be S3/GCS-backed, that's real egress cost. Can we add at least ETag/If-None-Match handling back?

Thanks for the feedback @secana, changes are made

secana:
Thanks for the quick update. Great addition!

PR #1366, opened 2026-09-04
feat(api): implement pagination for search

hishamhm:
This PR adds support for `page=` query argument when using the Registry Web API.

This implements a subset of the behavior of `page=` as seen in crates.io:

https://github.com/rust-lang/crates.io/blob/main/src/controllers/helpers/pagination.rs

Like crates.io, we add support for `page=` arguments, with 1-based counting. Pages are calculated based on the `per_page=` argument (i.e. with `page` and `per_page` we can determine `limit` and `offset`).

Unlike crates.io, we do not implement `seek`, since Kellnr results do not produce `seek` URLs.

For the `web-ui` API, in this PR we keep the same behavior we have previously; the `/search` endpoint does not implement pagination (it uses `offset = 0` and `limit = 100`).

### DIFF
--- a/crates/db/src/database/mod.rs
+++ b/crates/db/src/database/mod.rs
@@ -1392,9 +1392,12 @@ impl DbProvider for Database {
     async fn search_in_crate_name_and_description(
         &self,
         contains: &str,
+        limit: u64,
+        offset: u64,
         cache: bool,
     ) -> DbResult<Vec<CrateOverview>> {
-        self.query_crates(Some(contains), None, cache).await
+        self.query_crates(Some(contains), Some((limit, offset)), cache)
+            .await
     }
 
     async fn get_crate_overview_list(
--- a/crates/db/src/provider.rs
+++ b/crates/db/src/provider.rs
@@ -221,6 +221,8 @@ pub trait DbProvider: Send + Sync {
     async fn search_in_crate_name_and_description(
         &self,
         contains: &str,
+        limit: u64,
+        offset: u64,
         cache: bool,
     ) -> DbResult<Vec<CrateOverview>>;
     async fn get_crate_overview_list(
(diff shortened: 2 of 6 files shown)


secana:
Thanks again for the PR!

Three minor things I've found that I would like to see addressed before I merge:

1. `page` has no upper bound. `PerPage` is clamped to `0..=100`, but `Page::try_from` only rejects page < 1. `offset = (page - 1) * limit` can overflow usize for a large page value. That panics in debug builds andsilently wraps to a garbage offset in release. Could we bound page similarly to `per_page` (or at least guard the multiplication)?
2. `meta.total` doesn't reflect the real total match count. As far as I understand the crates.io API, `meta.total` should be the total number of matches across all pages, so clients know when to stop paginating. Right now it's set to `crates.len()` after `.take(per_page)`, so it'll basically always equal `per_page` until the last page. A spec-compliant client would think there's only one page and never request page 2+. Since this PR is what makes paging real, could we add a `COUNT(*)` (same filters, no LIMIT/OFFSET) so total is accurate?
3. The `.take(params.per_page.0)` in `search()` is now redundant, since the DB query already applies LIMIT/OFFSET. Could probably drop it.

hishamhm:
@secana Thank you for the review! I addressed all three items, please take a look.

For the `total`, I could not find a way to perform the `LIMIT/OFFSET` and full `COUNT(*)` in a single query [without caveats](https://stackoverflow.com/questions/28888375/run-a-query-with-a-limit-offset-and-also-get-the-total-number-of-rows) (or without using a nested query that would be functionally equivalent to two SELECTs), so I ended up refactoring the code to reuse the query construction and had it perform two queries. I think it ended up clean enough (and removed some of the `query`/`query2` repetition).

hishamhm:
@secana I finally got around to making `total` work correctly! Please re-review.

PR #1384, opened 2026-09-15
feat(auth): enforce SSO-only login with RP-initiated logout, vol 2

aergus-ferrous:
Hello, this rebases @hwittenborn's work in #1275 and attempts to address the points of https://github.com/kellnr/kellnr/pull/1275#pullrequestreview-4679925483. Here is a summary with open questions emphasized:

1. **Enforcement doesn't cover the registry API:** _What shared auth path should the SSO enforcement be checked in?_ In my (admittedly limited) analysis of the code, I couldn't find a single place where I could plug in an SSO enforcement check as it needs to know whether the endpoint that is being accessed interacts with passwords. (As a blunt fix, I added logic in `crates/auth/src/token.rs` that rejects password authentication when SSO is enforced.) 
2. **Total lockout when the IdP is unreachable at startup:** Made `run_server` exit with status code 1 if OAuth2 is supposed to be enforced but discovery fails. (I hope this is an appropriate place to add such logic.)
3. **Logout can become impossible with `auto_redirect`:** Encoded whether we come from a logout in a query and prevented redirection to the IDP if that is the case.
6. **The `!('error' in to.query)` escape hatch never fires:**  _What is the desired behavior for callback failures, specifically when `auto_redirect = true`?_ I have the impression that if we are in a state where kellnr enforces SSO but the callback fails consistently, this means that kellnr is basically unusable in that constellation and probably administrator intervention (e.g. fixing the IDP or the OAuth2 config) is required. One thing I can image is making the error more tangible for users by making the callback redirect to an error page instead of responding with a `4XX`, is that the direction you had in mind? (For now, I just removed the `!('error' in to.query)` condition.)
7. Smaller points: 
    1. **Cookie size limits:** _Were you thinking of something like extending `kellnr_db::SessionInfo` with a field (potentially) containing the OIDC ID token?_ I didn't do anything about token persistence (yet) because it sounded like it would have significant consequences all the way down to the database layer and might enlarge the scope of this PR too much.
    2. **Updating the end session endpoint upon rediscovery:** Implemented by putting the end session endpoint behind a mutex.
    3. **Password prompts in the UI when SSO is enforced:** Should not be visible anymore.
    4. **`ui/package-lock.json` diff**: Reverted.
    5. **UI tests for new login/logout flows**: Added.
    
 Edit: Sorry, had to rebase again after opening the PR due to recent dependency updates.


secana:
Thanks for picking this up! 

- You found the right hook. `Token::extract_token` is the single shared auth path and
   `Basic` is the only branch there that touches a password.
- An error page is the direction I had in mind, but not needed here. Worth noting though:
   `post_logout_redirect_uri` is `{base_url}/` with no marker, so the `from=logout` escape
   hatch only fires when the provider advertises no `end_session_endpoint`. Appending
   `?from=logout` would make both paths behave the same.
- Yes, extending `SessionInfo` is what I meant, and leaving it out was the right call.
   For sizing: after AES-GCM plus base64 a 3 KB ID token sits near the 4 KB cookie limit,
   so a size guard that skips the cookie and warns would be a cheap stopgap.

PR looks good to me!

